PR #9, opened 2016-11-16
Projections

prolic:


prolic:
Please carefully check the locking mechanisms!

src/Projection/MySQLEventStoreProjectionTrait.php
```diff
@@ -0,0 +1,62 @@
...
+use Prooph\EventStore\PDO\Exception\RuntimeException;
+
+trait MySQLEventStoreProjectionTrait
+{
+    /**
+     * @throws RuntimeException
+     */
+    protected function acquireLock(): void
+    {
+        $statement = $this->connection->prepare("SELECT GET_LOCK('$this->name', 1) as 'lock';");
```
lucascourot: no need to surround aliases with quotes

src/Projection/PostgresEventStoreProjectionTrait.php
```diff
@@ -0,0 +1,95 @@
...
+SELECT locked FROM $this->projectionsTable WHERE name = '$this->name';
+EOT;
+        $statement = $this->connection->prepare($sql);
+        $statement->execute();
+
+        $row = $statement->fetch(PDO::FETCH_OBJ);
+//die;
+        if (false === $row) {
+            $sql = <<<EOT
+INSERT INTO $this->projectionsTable (name, position, state, locked)
```
lucascourot: so you use GET_LOCK() for MySQL but a locked column for Postgres?

prolic: Yes. PostgreSQL has not the same feature.

src/Projection/MySQLEventStoreProjectionTrait.php
prolic: It's required!

src/Projection/PostgresEventStoreProjectionTrait.php
```diff
@@ -0,0 +1,95 @@
...
+    protected function acquireLock(): void
+    {
+        $sql = <<<EOT
+SELECT locked FROM $this->projectionsTable WHERE name = '$this->name';
+EOT;
+        $statement = $this->connection->prepare($sql);
+        $statement->execute();
+
+        $row = $statement->fetch(PDO::FETCH_OBJ);
+//die;
```
prolic: done

src/Projection/MySQLEventStoreProjectionTrait.php
lucascourot: why? can't you do simply ` as lock`?

prolic: No.

```
mysql> SELECT get_lock('ff', 3) as lock;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lock' at line 1
```

lucascourot: oh ok, fair enough, the 'lock' keyword is reserved :)

prolic:
damn, that's a bug. or maybe we do the same implementation as in psql?

Sascha-Oliver Prolic

2016-11-17 0:25 GMT+08:00 Lucas Courot <notifications@github.com>:

src/Projection/MySQLEventStoreProjectionTrait.php
prolic: damn, that's a bug. or maybe we do the same implementation as in psql?

prolic: I think having the same implementation in place might be the best scenario. Will update the code tomorrow.

lucascourot: Ok but don't forget to expose the timeout value in the config of the package so the user can configure it. I would actually be in favor of letting the user chose how he really wants to manage his locks by exposing a locking interface. Another solution would be to make your locked column a datetime meaning until when the projection is locked (instead of storing a random int).

prolic:
There is also a pthreads implementation planned for multi stream processing. Problem is different threads cannot share db connection. So how to construct this object? Inject a container and service names? Any ideas?

codeliner:
> There is also a pthreads implementation planned for multi stream processing. Problem is different threads cannot share db connection.

I followed @bendavies suggestion in the chat and took a look at icicle. I played a bit with reactPHP in the past and also read a few things about icicle.
Most of the examples are dealing with non-blocking streams but a google search took me there: https://packagist.org/packages/asyncphp/icicle-database
Except my experiments at home, I did not use non-blocking techniques in php so I cannot say much about it. Maybe we can ask the async php community for help or ideas ... non-blocking I/O works great for node.js and Go so why not also for php, but database connections would need to support it.

prolic:
- removed deadlocks
- added lock until timeout in milliseconds
- added projection factories
- added query factories
- added read model projection factories

src/Projection/PDOEventStoreProjectionTrait.php
```diff
@@ -0,0 +1,76 @@
...
+     * @throws RuntimeException
+     */
+    protected function acquireLock(): void
+    {
+        $now = new \DateTimeImmutable('now', new \DateTimeZone('UTC'));
+        $nowString = $now->format('Y-m-d\TH:i:s.u');
+        $lockUntilString = $now->modify('+' . (string) $this->lockTimeoutMs . ' ms')->format('Y-m-d\TH:i:s.u');
+
+        $sql = <<<EOT
+UPDATE $this->projectionsTable SET locked_until = ? WHERE name = ? AND (locked_until IS NuLL OR locked_until < ?);
```
bendavies: `NULL`

prolic: done

prolic:
About multi-stream-processing:

This will require pthreads and will probably be added with in a distinct implementation (much stuff differs then). Still will be added later and is not required at the moment.

prolic:
Note: This PR is now also cleaned up, so a DistributedLock interface can be added very easy (but is not yet done in this PR).

prolic:
Travis with postgres 9.4 hangs. But should be no problem, it's just travis :)

src/Projection/MySQLEventStoreProjection.php
```diff
@@ -0,0 +1,77 @@
...
+            $emitEnabled
+        );
+    }
+
+    public function delete(bool $deleteEmittedEvents): void
+    {
+        $this->connection->beginTransaction();
+
+        $listener = $this->eventStore->getActionEventEmitter()->attachListener(
+            ActionEventEmitterAware::EVENT_APPEND_TO,
```
codeliner: wrong event?

src/Projection/MySQLEventStoreReadModelProjection.php
```diff
@@ -0,0 +1,80 @@
...
+            $lockTimeoutMs
+        );
+    }
+
+    public function delete(bool $deleteEmittedEvents): void
+    {
+        $this->connection->beginTransaction();
+
+        $listener = $this->eventStore->getActionEventEmitter()->attachListener(
+            ActionEventEmitterAware::EVENT_APPEND_TO,
```
codeliner: wrong event?

prolic: removed

src/Projection/MySQLEventStoreProjection.php
prolic: fixed

PR #18, opened 2016-12-02
Persistence strategy

prolic:


prolic:
Question: Is Persistence Strategy the right name?

PR #28, opened 2017-01-02
Convert PostgreSQL payload column from JSONB to JSON.

### LINKED ISSUE #27: What is the reasoning behind using JSONB in PostgreSQL?
Question taken from the chat. Asked by @shochdoerfer

> Do you guys query the JSON payload a lot? If not the JSON field would be more efficient as no conversion happens.

shochdoerfer:
Fixes #27: Since the payload column simply stores JSON data it is more efficient to use the JSON datatype of PostgreSQL as no conversion happens when the data is getting stored or read.

### DIFF
--- a/src/PersistenceStrategy/PostgresAggregateStreamStrategy.php
+++ b/src/PersistenceStrategy/PostgresAggregateStreamStrategy.php
@@ -30,7 +30,7 @@ CREATE TABLE $tableName (
     no SERIAL,
     event_id CHAR(36) NOT NULL,
     event_name VARCHAR(100) NOT NULL,
-    payload JSONB NOT NULL,
+    payload JSON NOT NULL,
     metadata JSONB NOT NULL,
     created_at TIMESTAMP(6) NOT NULL,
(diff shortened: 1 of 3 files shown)


sandrokeil:
Seems something has been changed and is not up2date?

```
There were 2 warnings:
1) ProophTest\EventStore\PDO\Container\MySQLEventStoreFactoryTest::it_injects_plugins
Trying to configure method "setUp" which cannot be configured because it does not exist, has not been specified, is final, or is static

2) ProophTest\EventStore\PDO\Container\PostgresEventStoreFactoryTest::it_injects_plugins
Trying to configure method "setUp" which cannot be configured because it does not exist, has not been specified, is final, or is static
```

shochdoerfer:
The master build also fails with the same errors and warnings. Any pointers on how to fix this issue? Or should the fix be a separate PR?

prolic:
I take care of the failing tests. This is due to some changes on the plugins.

prolic:
thx @shochdoerfer, good catch!

PR #30, opened 2017-01-07
Projections Update

prolic:


src/Projection/PDOEventStoreQuery.php
```diff
@@ -0,0 +1,392 @@
...
+    /**
+     * @var int lock timeout in milliseconds
+     */
+    private $lockTimeoutMs;
+
+    /**
+     * @var int
+     */
+    private $eventCounter = 0;
+
```
sandrokeil: `$lockTimeoutMs` and `$eventCounter` is never used!?

src/Projection/PDOEventStoreProjectionFactory.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\PDO\Projection;
+
+use PDO;
```
sandrokeil: `use PDO` is never used. Please remove

sandrokeil:
The parameter type description here `\Prooph\EventStore\PDO\PDOStreamIterator::$currentItem` is wrong. `array|false` and access with `$this->currentItem->created_at`?

sandrokeil:
Using always camel case looks better to me e.g. `PDO` -> `Pdo` or `MySQL` -> `MySql`. Because `PDOEventStoreException` is *hard* to read. What do you think?

src/Projection/PDOEventStoreProjectionFactory.php
prolic: it's used in the constructor

prolic:
@sandrokeil updated

prolic:
@codeliner @sandrokeil @oqq about naming:

I am not sure about the last commit, the base class in PHP is PDO, not Pdo. Also Doctrine uses this f.e. https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySQL57Platform.php

oqq:
There are several comments to change this in php code and also in doctrine. Especially in doctrine code, cuz the PDO prefix not fits to PSR-1.

I am for Pdo instead of PDO since 1 year for that PSR reasons and have used PDO before. But it´s a matter of taste, so no hard opinion from me.

bweston92:
Can it be explained why we are coupling a read model with the event store?

prolic:
@bweston92 I don't get your question. The read model is a projection of the recorded events. So in order to create a readmodel, you have to get a stream of events and process it. Nothing else. The readmodel itself has no dependency on the event store. The projector needs to have knowledge about the event store and the read model, in order to create it.

bweston92:
@prolic sorry I just seen read model and projection key words in the event store.

PR #33, opened 2017-01-10
adds missing phpunit group annotations to container test classes

oqq:
UnitTests would fail with messages like

> "Extension "pdo_pgsql" is not loaded"


prolic:
Can you also put the factories to the excluded files for coverage in the depending phpunit files?

PR #41, opened 2017-01-13
Always call reset projection

### LINKED ISSUE #38: Should resetProjection always be called?
I'm wondering if [this line](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreReadModelProjection.php#L327) should be moved above the [if block](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreReadModelProjection.php#L315) so that `$this->resetProjection()` gets called even if the initCallback returns a state array?

codeliner:
Fixes #38


codeliner:
Decreased coverage can be ignored

PR #44, opened 2017-01-18
use prepared load statements

oqq:
Currently the queries in aggregate load methods from event stores builds queries and recreate this for batch requests.

This PR changes this behavior and implements prepared statements. The PdoStreamIterator has only to change the values for `fromNumber` and `limit` without have to know how to queries should be builded.

### DIFF
--- a/src/MySqlEventStore.php
+++ b/src/MySqlEventStore.php
@@ -283,7 +283,7 @@ EOT;
             $values[$parameter] = $value;
         }
 
-        $where[] = "`no` >= :fromNumber";
+        $where[] = '`no` >= :fromNumber';
 
         $whereCondition = implode(' AND ', $where);
         $limit = min($count, $this->loadBatchSize);
@@ -358,7 +358,7 @@ EOT;
             $values[$parameter] = $value;
         }
 
-        $where[] = "`no` <= :fromNumber";
+        $where[] = '`no` <= :fromNumber';
 
         $whereCondition = implode(' AND ', $where);
         $limit = min($count, $this->loadBatchSize);
--- a/src/MySqlEventStore.php
+++ b/src/MySqlEventStore.php
@@ -265,40 +265,44 @@ EOT;
 
         $tableName = $this->persistenceStrategy->generateTableName($streamName);
 
-        $sql = [
-            'from' => "SELECT * FROM $tableName",
-            'orderBy' => 'ORDER BY no ASC',
-        ];
+        $where = [];
+        $values = [];
 
-        foreach ($metadataMatcher->data() as $match) {
+        foreach ($metadataMatcher->data() as $key => $match) {
             $field = $match['field'];
             $operator = $match['operator']->getValue();
             $value = $match['value'];
+            $parameter = ':metadata_'.$key;
 
             if (is_bool($value)) {
-                $value = var_export($value, true);
-            } elseif (is_string($value)) {
-                $value = $this->connection->quote($value);
+                $where[] = "metadata->\"$.$field\" $operator ".var_export($value, true);
+                continue;
             }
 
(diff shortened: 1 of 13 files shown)


src/PdoStreamIterator.php
```diff
@@ -78,7 +76,6 @@ public function __construct(
         PDO $connection,
         PDOStatement $statement,
         MessageFactory $messageFactory,
-        array $sql,
         ?int $batchSize,
         int $fromNumber,
         int $count,
```
prolic: is `bool forward` still needed then?

src/PdoStreamIterator.php
```diff
@@ -196,25 +191,15 @@ public function rewind(): void
...
     }
 
-    private function buildStatement(array $sql, int $fromNumber): PDOStatement
+    private function buildStatement(int $fromNumber): PDOStatement
     {
         $limit = $this->count < ($this->batchSize * ($this->batchPosition + 1))
             ? $this->count - ($this->batchSize * $this->batchPosition)
             : $this->batchSize;
 
-        if ($this->forward) {
```
prolic: see you removed `forward`

src/PostgresEventStore.php
```diff
@@ -308,42 +309,43 @@ public function loadReverse(
...
 
             if (is_bool($value)) {
-                $value = var_export($value, true);
-                $sql['where'][] = "metadata ->> '$field' $operator '$value'";
-            } elseif (is_string($value)) {
-                $value = $this->connection->quote($value);
-                $sql['where'][] = "metadata ->> '$field' $operator $value";
-            } else {
-                $sql['where'][] = "metadata ->> '$field' $operator '$value'";
+                $where[] = "metadata->'$field' $operator '".var_export($value, true)."'";
```
prolic: `metadata ->>` instead of `metadata -> `

see: https://www.postgresql.org/docs/9.3/static/functions-json.html

`->` means: Get JSON object field
`->>` means: Get JSON object field as text

src/MySqlEventStore.php
```diff
@@ -337,40 +340,43 @@ public function loadReverse(
...
             $field = $match['field'];
             $operator = $match['operator']->getValue();
             $value = $match['value'];
+            $parameter = ':metadata_'.$key;
 
             if (is_bool($value)) {
-                $value = var_export($value, true);
-            } elseif (is_string($value)) {
-                $value = $this->connection->quote($value);
+                $where[] = "metadata->\"$.$field\" $operator ".var_export($value, true);
```
prolic: I think this should be `metadata->>` to unquote.

src/PdoStreamIterator.php
oqq: Is still needed here: https://github.com/oqq/pdo-event-store/blob/82a55d82a6bf93f20ecc0941198a608c22320590/src/PdoStreamIterator.php#L139

src/PostgresEventStore.php
oqq: I am not sure about that, but mysql declares 

`->>` as shortcut for JSON_UNQUOTE(JSON_EXTRACT()), and
`->` as shortcut for JSON_EXTRACT()

So `->>` is always an extra step. With `->` all tests passes. Maybe Postgre is fine with. Should we add an extra step?

src/MySqlEventStore.php
oqq: Than it would also match 'true' === true. Right?

src/PdoStreamIterator.php
prolic: ah ok

src/PostgresEventStore.php
prolic: no, ok when it works

prolic:
great work! thx! :+1:

PR #49, opened 2017-02-11
Remove usage of MessageConverter

oqq:
After investigate the code I have found out the MessageConverter is not in use in booth event stores.
I am not sure if this should be used again or all usages could be removed.

But less is more, so I have removed it.


prolic:
manually merged into master

Issue #50, opened 2017-02-12
Add possibility to send messages to projectors


prolic:
Basic idea:

Send a message to a projector, that it should stop / delete / reset

PR #51, opened 2017-02-12
delete/reset/stop projections

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/50


prolic:
this one is really nice for event-store-http-api !!!

src/MySqlEventStore.php
```diff
@@ -490,13 +499,59 @@ public function getDefaultReadModelProjectionFactory(): ReadModelProjectionFacto
...
         }
 
         return $this->defaultReadModelProjectionFactory;
     }
 
+    public function deleteProjection(string $name, bool $deleteEmittedEvents): void
+    {
+        $sql = <<<EOT
+UPDATE $this->projectionsTable SET status = ? WHERE name = ? LIMIT 1;
+EOT;
```
sandrokeil: Why is `LIMIT 1` here used and not in the `PostgresEventStore` class?

src/Projection/PdoEventStoreReadModelProjection.php
```diff
@@ -368,19 +407,37 @@ public function run(bool $keepRunning = true): void
...
+
+                return;
+            case ProjectionStatus::DELETING_INCL_EMITTED_EVENTS():
+                $this->delete(true);
+
+                return;
+            case ProjectionStatus::RESETTING():
+                $this->reset();
+                break;
+        }
```
sandrokeil: `default` case should be added, also if nothing is to do.

src/MySqlEventStore.php
prolic: Postgres has not UPDATE ... LIMIT. But we can remove it from MySqlEventStore for consistence reasons, anyway, there no possibility to update more than one row, as name is unique.

src/Projection/PdoEventStoreReadModelProjection.php
prolic: done

src/MySqlEventStore.php
sandrokeil: Alright, let it as is. I was just wondering, because of my zero knowledge about postgres.

codeliner:
that is really cool stuff. The event-store-ui will be a killer feature!

PR #52, opened 2017-02-15
fetchStreamNames, fetchCategoryNames, fetchProjectionNames

prolic:


src/MySqlEventStore.php
```diff
@@ -552,11 +512,166 @@ public function stopProjection(string $name): void
...
+
+        $where = [];
+        $values = [];
+
+        if (null !== $filter && $regex) {
+            $where[] = '`real_stream_name` REGEXP :filter ';
+            $values[':filter'] = $filter;
+        } elseif (null !== $filter && ! $regex) {
+            $where[] = '`real_stream_name` = :filter ';
+            $values[':filter'] = $filter;
```
prolic: only exact match! should be change that to `LIKE %filter%` ?

If so, this has to be changed at InMemoryEventStore as well

src/MySqlEventStore.php
```diff
@@ -552,11 +512,166 @@ public function stopProjection(string $name): void
...
+        if ($regex && false === @preg_match("/$filter/", '')) {
+            throw new Exception\InvalidArgumentException('Invalid regex pattern given');
+        }
+
+        $where = [];
+        $values = [];
+
+        if (null !== $filter && $regex) {
+            $where[] = '`real_stream_name` REGEXP :filter ';
+            $values[':filter'] = $filter . '.*-.+';
```
prolic: regex ok?

src/MySqlEventStore.php
```diff
@@ -552,11 +512,166 @@ public function stopProjection(string $name): void
...
+        } elseif (null !== $filter && ! $regex) {
+            $where[] = '`real_stream_name` LIKE :filter ';
+            $values[':filter'] = $filter . '-%';
```
prolic: again, only exact match

src/MySqlEventStore.php
```diff
@@ -552,11 +512,166 @@ public function stopProjection(string $name): void
...
+
+        $whereCondition = implode(' AND ', $where);
+        if (! empty($whereCondition)) {
+            $whereCondition = 'WHERE ' . $whereCondition;
+        }
+
+        $query = <<<SQL
+SELECT `real_stream_name` FROM $this->eventStreamsTable
+$whereCondition
+ORDER BY `no`
```
prolic: what about order? by real_stream_name instead? If so, this has to be changed at InMemoryEventStore as well.

src/MySqlEventStore.php
```diff
@@ -552,11 +512,166 @@ public function stopProjection(string $name): void
...
+
+        $where[] = 'substring(real_stream_name, 1, locate(\'-\', real_stream_name) - 1) != \'\'';
+
+        $whereCondition = 'WHERE ' . implode(' AND ', $where);
+
+        $query = <<<SQL
+SELECT substring(real_stream_name, 1, locate('-', real_stream_name) - 1) as `category` FROM $this->eventStreamsTable
+$whereCondition
+GROUP BY `category`
+LIMIT $offset, $limit
```
prolic: No order at all currently! Important?

src/MySqlEventStore.php
basz: seem ok : https://regex101.com/r/ykVagg/1

prolic: well the regex is a valid regex, but if it's good for the task or should be altered is the question :)

prolic: do we want this https://regex101.com/r/ykVagg/2 ?

basz: but i can only guess at what task it is you try to do? Reading the regex I would say you try to select '[FILTER]ANYTHING-ANYTHING'. Why would you want to alter it? Are you looking for a more efficient version

src/MySqlEventStore.php
sandrokeil: In MySQL, if you start with a `%` I guess no index is used.

src/MySqlEventStore.php
sandrokeil: We should consider if the index can be used for sorting. I think `ORDER BY real_stream_name` is slower than `ORDER BY no`.

src/MySqlEventStore.php
sandrokeil: If we need no order we shoud add `ORDER BY NULL`. This is a trick to avoid filesystem / temp table sorts. But we should check what MySQL `explain` said.

src/MySqlEventStore.php
```diff
@@ -263,35 +263,15 @@ public function load(
...
-
-            if (is_bool($value)) {
-                $where[] = "metadata->\"$.$field\" $operator ".var_export($value, true);
-                continue;
-            }
-
-            $where[] = "metadata->\"$.$field\" $operator $parameter";
-            $values[$parameter] = $value;
-        }
+        list($where, $values) = $this->createWhereClauseForMetadata($metadataMatcher);
```
malukenho: `[$where, $values] = $this->createWhereClauseForMetadata($metadataMatcher);`

bweston92:
LGTM.

PR #56, opened 2017-02-17
provides default dsn value for charset in factories

oqq:
If anyone trusts the default config (e.g. me), than this person could get bad experience with the mysql event store without this PR.

json_decode() would return null with an invalid json string, e.g. a not correct utf-8 encoded string.
Which would end up with `Assert\InvalidArgumentException: payload must be an array`.

Anyway.. if there is no reason why the default charset should not be utf-8, this PR should be merged before next stable release to prevent for bc breaks.


prolic:
wrong for postgres - build fails

prolic:
must be something like: `pg_connect("host=localhost options='--client_encoding=UTF8'");`

oqq:
how about to provide the dsn from the entire factory instead of in the abstract one?
Maybe a getDsn method could help. Meanings?

oqq:
Here is how doctrine get around this..
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php#L56

oqq:
I have added my proposal and have extended tests. Please review again @prolic .

src/Container/PostgresEventStoreFactory.php
```diff
@@ -52,16 +52,35 @@ protected function eventStoreClassName(): string
...
+        }
+
+        return $dsn;
+    }
+
     public function defaultOptions(): iterable
     {
         return [
             'connection_options' => [
-                'driver' => 'pdo_pgsql',
```
prolic: Why is that removed?

tests/TestUtil.php
```diff
@@ -45,6 +45,7 @@ public static function getConnection(): PDO
             $dsn .= 'dbname=' . $connectionParams['dbname'] . $separator;
             $dsn = rtrim($dsn);
             self::$connection = new PDO($dsn, $connectionParams['user'], $connectionParams['password']);
+            self::$connection->query("SET NAMES '".$connectionParams['charset']."'");
```
prolic: Why not via connection options?

src/Container/AbstractEventStoreFactory.php
```diff
@@ -94,15 +81,15 @@ public function __invoke(ContainerInterface $container): EventStore
...
-            $password = $config['connection_options']['password'];
-            $connection = new PDO($dsn, $user, $password);
+            $connection = new PDO(
+                $this->buildConnectionDsn($config['connection_options']),
+                $config['connection_options']['user'],
+                $config['connection_options']['password']
+            );
+
+            if (isset($config['connection_options']['charset'])) {
+                $connection->query("SET NAMES '".$config['connection_options']['charset']."'");
```
prolic: Why not use connection options?

src/Container/PostgresEventStoreFactory.php
oqq: Is not needed anymore.

tests/TestUtil.php
oqq: I have not found a connection option which would work for postgre. Thats the way how doctrine is bypass this issue.

src/Container/AbstractEventStoreFactory.php
oqq: The connection option for charset is used in mysql factory. This is a workaround for postgre.

prolic:
Try options='--client_encoding=UTF8' for PostgreSQL

oqq:
@prolic any reason why you have used a space as separator for postgre? It seems that postgre is fine with ';' as separator.

prolic:
| @prolic any reason why you have used a space as separator for postgre? It seems that postgre is fine with ';' as separator.

As far as I know ; doesn't work for postgres.

Issue #57, opened 2017-02-17
PersistBlockSize projection option works not properly


oqq:
If i don´t handle all events in a read model projection, the persistBlockSize option value could be bypassed and therefore the persist procedure would never be called until the end of the script.

There are 2 simple solutions for that issue:
- run `$this->eventCounter++;` after `continue` from `handleStreamWithHandlers` method
- change the expression to `$this->eventCounter >= $this->persistBlockSize`

Could provide a PR. Which is the preferred solution @prolic ?

PR #58, opened 2017-02-18
PersistBlockSize projection option works not properly

prolic:
fixes https://github.com/prooph/pdo-event-store/issues/57

/cc @oqq


oqq:
Maybe a unit test would be fine. ;)

prolic:
@oqq Go for it!

PR #69, opened 2017-03-26
remove lock timeout option (added to event-store interface)

prolic:
depends on https://github.com/prooph/event-store/pull/280


prolic:
Travis fails, because it used PHP 7.1.3!

PR #70, opened 2017-03-26
conflict with php 7.1.3

prolic:
@sandrokeil like this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Requirements
 For MySQL you need server version >= 5.7.9.
 For Postgres you need server version >= 9.4.
 
+Due to a bug in PHP 7.1.3 this library is not compatible with that specific php version.
+
 Setup
 -----
 
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "ext-pdo_pgsql": "For usage with PostgreSQL"
     },
     "conflict": {
+        "php": "7.1.3",
         "sandrokeil/interop-config": "<2.0.1"
     },
     "autoload": {


basz:
Q: why only on the store? This DateBug would influence every time stamp on every event in de bus as well. Whether persisted or not it seems to me a wider problem then just the pdo store?

prolic:
Good question, the event store tests at least do pass, the pdo-event-store tests don't. That's why I thought it's only related to this component. Should we add this to prooph/common (because this one is used everywhere) or to every single component?

Ideas?

/cc @codeliner @sandrokeil

basz:
disallow completely (common seems a good option) as I consider 7.1.3 DOA, at least for these kind of applications...

prolic:
ping @codeliner @sandrokeil use prooph/common only?

codeliner:
:+1: for prooph/common that is definitely the right place for an issue that effects alle components

sandrokeil:
@prolic @basz Is this only affected if we modify the Datetime like in the [example](https://3v4l.org/I2fQq)?

`$lockUntilString = $now->modify('+' . (string) 1000 . ' ms')->format('Y-m-d\TH:i:s.u');`

So I guess this is an edge case and we should do nothing, because we don't modify the timestamp in the prooph components. Otherwise we should put it only in prooph\common like @basz mentioned.

basz:
I don't think so. It's a regression bug to php5 behaviour (no microtime)
 
https://3v4l.org/YeJTe

prolic:
done in prooph/common and released 4.1, see https://github.com/prooph/common/releases/tag/v4.1.0

PR #71, opened 2017-03-29
remove uniqueViolationErrorCodes from persistance strategy + add docs

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/31

/cc @basz @oqq


prolic:
Travis build failed because of the datetime bug in PHP 7.1.3, you can ignore that.

PR #75, opened 2017-05-09
add transaction support to mysql

iampersistent:
I'm not thrilled with the workaround for the mysql implicit commit, but it does the job. After I came up with this solution, I considered using a TEMPORARY TABLE but didn't pursue it. 

If the consensus is this solution smells too much, I can look at potentially using the temp table.

### DIFF
--- a/src/MySqlEventStore.php
+++ b/src/MySqlEventStore.php
@@ -20,14 +20,17 @@ use Prooph\EventStore\EventStore;
 use Prooph\EventStore\Exception\ConcurrencyException;
 use Prooph\EventStore\Exception\StreamExistsAlready;
 use Prooph\EventStore\Exception\StreamNotFound;
+use Prooph\EventStore\Exception\TransactionAlreadyStarted;
+use Prooph\EventStore\Exception\TransactionNotStarted;
 use Prooph\EventStore\Metadata\MetadataMatcher;
 use Prooph\EventStore\Pdo\Exception\ExtensionNotLoaded;
 use Prooph\EventStore\Pdo\Exception\RuntimeException;
 use Prooph\EventStore\Stream;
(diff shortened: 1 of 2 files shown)


prolic:
The problem is, that it's _NOT_ a transaction here.
After `CREATE TABLE` mysql will automatically commit the transaction, so there is no transaction running after that.

see: https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html

A temporary table might help, but i'm not sure how to make it non-temporary afterwards within this transaction.

iampersistent:
I was thinking with the temp table, track what table needed to be created post-commit, then create the new table and copy the data from the temp table over. Its still a bit of hack.

The real problem is that I am trying to use the `TransactionManager` in the event store bus bridge with the `MySqlEventStore`. Perhaps I shouldn't be doing that. TBH, I was blindly setting things up to get them to work, trying to follow the examples and docs, and I used that plugin.

prolic:
Don't use the transaction manager with MySQL, because it doesn't support
transactions. Simple ;-)

On May 15, 2017 01:37, "Shank" <notifications@github.com> wrote:

I was thinking with the temp table, track what table needed to be created
post-commit, then create the new table and copy the data from the temp
table over. Its still a bit of hack.

The real problem is that I am trying to use the TransactionManager in the
event store bus bridge with the MySqlEventStore. Perhaps I shouldn't be
doing that. TBH, I was blindly setting things up to get them to work,
trying to follow the examples and docs, and I used that plugin.

—
You are receiving this because you were assigned.

Reply to this email directly, view it on GitHub
<https://github.com/prooph/pdo-event-store/pull/75#issuecomment-301327611>,
or mute the thread

iampersistent:
I agree. Closing.

PR #76, opened 2017-05-09
catch exception to ignore

iampersistent:


src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -595,8 +596,11 @@ private function createProjection(): void
 EOT;
 
         $statement = $this->connection->prepare($sql);
-        // we ignore any occuring error here (duplicate projection)
-        $statement->execute([$this->name, $this->status->getValue()]);
+        try {
+            $statement->execute([$this->name, $this->status->getValue()]);
+        } catch (PDOException $e) {
```
prolic: there is _no_ exception happening. unique constraint is _no_ exception.

iampersistent: The comment suggested to me that it was supposed to fall through. I assumed it was set up this way to save a SELECT to see if it exists.

However, I think I need to add a check so that only error 1062 passes through.

prolic: Ok, sounds good, but don't catch exception then, check the error code on the statement object instead.

About the second PR, I will take a look this weekend.

prolic:
ping @iampersistent still working on this one?

src/Projection/PdoEventStoreReadModelProjector.php
dragosprotung: I actually wanted to solve the same issue. The problem with checking error codes is that you get different ones depending on the driver. 
```php
[
   "pgsql" => "23505",
   "mysql" => "1062"
]
```

I could make a new PR with this in mind (I don't think I can add all the error codes for all the drivers, but new PR could solve this). @prolic Would you accept this solution ?

prolic:
Error codes for mysql and postgres should be enough. Everything else is not
tested anyway.

dragosprotung:
Then I will come up with new PR in the next hour(s)

iampersistent:
@prolic sorry, I've been covered up. But that's no excuse, I should have taken a minute to reply. 

Thanks @dragosprotung for handling this.

PR #77, opened 2017-05-16
Catch and ignore duplicate projection error when creating it

dragosprotung:
This is an attempt to fix #76 

Unfortunately configuring the PDO connection from the beginning to throw exceptions on errors will result in almost all tests failing.
I think this should be addressed, but I believe it's outside of the scope of current issue.

@prolic any suggestions on how to make this better ?

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -24,6 +24,7 @@ use Prooph\Common\Messaging\Message;
 use Prooph\EventStore\EventStore;
 use Prooph\EventStore\EventStoreDecorator;
 use Prooph\EventStore\Exception;
+use Prooph\EventStore\Pdo\Exception\ProjectionNotCreatedException;
 use Prooph\EventStore\Pdo\MySqlEventStore;
 use Prooph\EventStore\Pdo\PostgresEventStore;
 use Prooph\EventStore\Projection\ProjectionStatus;
@@ -34,6 +35,11 @@ use Prooph\EventStore\Util\ArrayCache;
 
 final class PdoEventStoreProjector implements Projector
 {
+    private const UNIQUE_VIOLATION_ERROR_CODES = [
+        'pgsql' => '23505',
+        'mysql' => '23000',
+    ];
+
     /**
      * @var EventStore
      */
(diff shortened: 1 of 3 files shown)


src/Projection/PdoEventStoreProjector.php
```diff
@@ -633,8 +638,15 @@ private function createProjection(): void
 EOT;
 
         $statement = $this->connection->prepare($sql);
-        // we ignore any occuring error here (duplicate projection)
-        $statement->execute([$this->name, $this->status->getValue()]);
+        try {
+            $statement->execute([$this->name, $this->status->getValue()]);
+        } catch (\PDOException $exception) {
```
prolic: As you noticed already, this lib expects currently ERRMODE_SILENT, which is the default anyway. So I think we should remove the catching of exception here and only check for errorCode on the statement after query.

src/Projection/PdoEventStoreProjector.php
```diff
@@ -34,6 +34,11 @@
 
 final class PdoEventStoreProjector implements Projector
 {
+    private const UNIQUE_VIOLATION_ERROR_CODES = [
```
prolic: nice :+1:

src/Projection/PdoEventStoreProjector.php
prolic: Also let's throw a custom exception, like `ProjectionNotCreated`

prolic:
About the Pdo::ATTR_ERRMODE, what do you think if we add this to the documentation and set it to ERRMODE_SILENT in the contructor of the event-store, projection, read model projection and query classes?

src/Projection/PdoEventStoreProjector.php
dragosprotung: I am now expecting silent mode and throw a custom exception

dragosprotung:
Changing the error mode will solve my problem :) but I am not sure it's the best way to go. 
I think having a solution that keeps the behavior the same regardless of how the PDO connection is configured is a better approach.

dragosprotung:
There are a lot of things that will break if exceptions are thrown.
I am willing to work on a patch for this, but in order to keep BC we need to wrap every call in a try/catch block and if no exception is thrown then call some kind of helper class to detect specific error codes and either ignore them or throw custom exceptions.

What would you suggest ?

prolic:
What do you think about try-catch-ignore all PDOExceptions?

dragosprotung:
I guess that would be a good compromise. Shall I create a PR with this approach ?

prolic:
Nice. Thanks

PR #82, opened 2017-05-30
add event position if field not occupied

prolic:


prolic:
@fieg we cannot rename the db field at this time as this would be a BC break and requires a new major version.

prolic:
@fieg If you want, please add some documentation in this repo somewhere covering this fact.

PR #89, opened 2017-06-21
Fix path to phpunit files from composer.json

bweston92:
Unable to run the tests due to file names not matching in composer.json to the actual file names.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,8 @@
     "scripts": {
         "cs": "php-cs-fixer fix -v --diff --dry-run",
         "cs-fix": "php-cs-fixer fix -v --diff",
-        "test-postgres": "vendor/bin/phpunit -c phpunit.xml.postgres",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -63,7 +63,7 @@
     "scripts": {
         "cs": "php-cs-fixer fix -v --diff --dry-run",
         "cs-fix": "php-cs-fixer fix -v --diff",
-        "test-postgres": "vendor/bin/phpunit -c phpunit.xml.pdo_pgsql",
-        "test-mysql": "vendor/bin/phpunit -c phpunit.xml.pdo_mysql"
+        "test-postgres": "vendor/bin/phpunit -c phpunit.xml.postgres",
+        "test-mysql": "vendor/bin/phpunit -c phpunit.xml.mysql"
```
prolic: can you add test-mariadb as well?

PR #91, opened 2017-06-21
Added test for MySqlSimpleStreamStrategy

bweston92:


tests/PersistenceStrategy/MySqlSimpleStreamStrategyTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $events = new ArrayIterator([
+            new class($expected) extends DomainEvent {
+                use PayloadTrait;
+            }
+        ]);
+
+        $out = $stategy->prepareData($events);
+
+        $this->assertEquals($expected, json_decode($out[2], true));
```
afoeder: I think I would go with https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertJsonStringEqualsJsonString and an actual string comparison; it's what we want here: the json STRING being as expected.

prolic:
Obviously travis passed without any fix applied.

tests/PersistenceStrategy/MySqlSimpleStreamStrategyTest.php
bweston92: Well no because that means we would encode with the same flags. We no the flags harm the string but we need to test against the impact.

PR #92, opened 2017-06-21
Documentation update for running tests.

bweston92:
I think it may be favourable to have the Docker way work without configuration example environment flags for the sacrifice using composer on host machine.

The way around this is we could revert what I did previously and use a `docker_tests.sh` which will set the correct environment variable, wait for the database to be available and execute the test.


codeliner:
If this is the easiest way to get around the issue then I can live with it.

PR #95, opened 2017-06-28
update lock after sleep

prolic:
/cc @fritz-gerneth can you test if this works for you?


fritz-gerneth:
Can confirm the lock gets updated while the projection is running idle (after your second commit :)).

PR #96, opened 2017-06-29
Optionally trigger pcntl_signal_dispatch

fritz-gerneth:
Currently there is no way to gracefully shut down a running projection when any OS signals are received. Setting the projection is `stopping` state via a custom CLI script is not always an option (e.g. when projections and others are managed by supervisord). 

I think how the different signals should be handled should be left to the programmer, I have included a short example for reference below. In any case, to support this feature the projection loop at least needs to trigger the signal dispatch function.

Some remarks on my approach:
- `declare(ticks=1)` is not required anymore to my knowledge, it got replaced by `pcntl_signal_dispatch`
- I made this argument optional, defaulting to `false` to remain fully backwards compatible
- The location of the `triggerPcntlSignalDispatch` call within the loop has quite a few implications. It might even make sense to trigger this before `usleep`. 

```
        $pcntlCallback = function () use ($name) {
            echo "Attempting graceful shutdown of projection " . $name . PHP_EOL;
            $this->currentProjection->stop();
        };
        pcntl_signal(SIGTERM, $pcntlCallback);
        pcntl_signal(SIGINT, $pcntlCallback);
        pcntl_signal(SIGQUIT, $pcntlCallback);

      $this->currentProjection->start();
```

### DIFF
--- a/src/Projection/MariaDbProjectionManager.php
+++ b/src/Projection/MariaDbProjectionManager.php
@@ -88,7 +88,7 @@ final class MariaDbProjectionManager implements ProjectionManager
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
             $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
-            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
+            $options[PdoEventStoreProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
--- a/src/Projection/MariaDbProjectionManager.php
+++ b/src/Projection/MariaDbProjectionManager.php
@@ -87,7 +87,8 @@ final class MariaDbProjectionManager implements ProjectionManager
             $options[PdoEventStoreProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
@@ -105,7 +106,8 @@ final class MariaDbProjectionManager implements ProjectionManager
             $this->projectionsTable,
             $options[PdoEventStoreReadModelProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreReadModelProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreReadModelProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreReadModelProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreReadModelProjector::OPTION_SLEEP] ?? PdoEventStoreReadModelProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreReadModelProjector::OPTION_SLEEP] ?? PdoEventStoreReadModelProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
--- a/src/Projection/MySqlProjectionManager.php
+++ b/src/Projection/MySqlProjectionManager.php
@@ -87,7 +87,8 @@ final class MySqlProjectionManager implements ProjectionManager
             $options[PdoEventStoreProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
@@ -105,7 +106,8 @@ final class MySqlProjectionManager implements ProjectionManager
             $this->projectionsTable,
             $options[PdoEventStoreReadModelProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreReadModelProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreReadModelProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreReadModelProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreReadModelProjector::OPTION_SLEEP] ?? PdoEventStoreReadModelProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreReadModelProjector::OPTION_SLEEP] ?? PdoEventStoreReadModelProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
--- a/src/Projection/MySqlProjectionManager.php
+++ b/src/Projection/MySqlProjectionManager.php
@@ -88,7 +88,7 @@ final class MySqlProjectionManager implements ProjectionManager
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
             $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
-            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
+            $options[PdoEventStoreProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreProjector::DEFAULT_PCNTL_DISPATCH
         );
     }
 
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -134,6 +134,11 @@ final class PdoEventStoreProjector implements Projector
     private $sleep;
 
     /**
+     * @var bool
+     */
+    private $triggerPcntlSignalDispatch;
+
+    /**
      * @var array|null
      */
     private $query;
@@ -152,7 +157,8 @@ final class PdoEventStoreProjector implements Projector
         int $lockTimeoutMs,
         int $cacheSize,
         int $persistBlockSize,
-        int $sleep
+        int $sleep,
+        bool $triggerPcntlSignalDispatch = false
     ) {
         $this->eventStore = $eventStore;
         $this->connection = $connection;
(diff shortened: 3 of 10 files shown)


src/Projection/PdoEventStoreProjector.php
```diff
@@ -164,6 +170,7 @@ public function __construct(
         $this->persistBlockSize = $persistBlockSize;
         $this->sleep = $sleep;
         $this->status = ProjectionStatus::IDLE();
+        $this->triggerPcntlSignalDispatch = extension_loaded('pcntl') && $triggerPcntlSignalDispatch;
```
codeliner: switch the order of the checks so that if `$triggerPcntlSignalDispatch = false` PHP does not check if extension is loaded

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -151,6 +161,7 @@ public function __construct(
         $this->persistBlockSize = $persistBlockSize;
         $this->sleep = $sleep;
         $this->status = ProjectionStatus::IDLE();
+        $this->triggerPcntlSignalDispatch = extension_loaded('pcntl') && $triggerPcntlSignalDispatch;
```
codeliner: same as above

codeliner:
@prolic can you take a look? You are the expert in that area.

Also a test and docs update would be nice, especially regarding $pcntlCallback

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -35,6 +35,10 @@
 
 final class PdoEventStoreReadModelProjector implements ReadModelProjector
 {
+    const OPTION_PCNTL_DISPATCH = 'trigger_pcntl_dispatch';
```
prolic: public const

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -35,6 +35,10 @@
...
+
+    const DEFAULT_PCNTL_DISPATCH = false;
```
prolic: public const

src/Projection/PdoEventStoreProjector.php
```diff
@@ -134,6 +134,11 @@
     private $sleep;
 
     /**
+     * @var bool
+     */
+    private $triggerPcntlSignalDispatch;
+
+    /**
```
prolic: new consts missing

src/Projection/MySqlProjectionManager.php
```diff
@@ -87,7 +87,8 @@ public function createProjection(
             $options[PdoEventStoreProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
```
prolic: wrong cost usage

src/Projection/MariaDbProjectionManager.php
```diff
@@ -87,7 +87,8 @@ public function createProjection(
             $options[PdoEventStoreProjector::OPTION_LOCK_TIMEOUT_MS] ?? PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
```
prolic: wrong const usage

src/Projection/PostgresProjectionManager.php
```diff
@@ -87,7 +87,8 @@ public function createProjection(
             $options[PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS] ?? PdoEventStoreProjector::DEFAULT_LOCK_TIMEOUT_MS,
             $options[PdoEventStoreProjector::OPTION_CACHE_SIZE] ?? PdoEventStoreProjector::DEFAULT_CACHE_SIZE,
             $options[PdoEventStoreProjector::OPTION_PERSIST_BLOCK_SIZE] ?? PdoEventStoreProjector::DEFAULT_PERSIST_BLOCK_SIZE,
-            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP
+            $options[PdoEventStoreProjector::OPTION_SLEEP] ?? PdoEventStoreProjector::DEFAULT_SLEEP,
+            $options[PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH] ?? PdoEventStoreReadModelProjector::DEFAULT_PCNTL_DISPATCH
```
prolic: wrong const usage

src/Projection/PdoEventStoreProjector.php
```diff
@@ -498,6 +505,8 @@ public function run(bool $keepRunning = true): void
 
                 $this->eventCounter = 0;
 
+                $this->triggerPcntlSignalDispatch();
```
prolic: I think this might be the best place for that function call

prolic:
@fritz-gerneth do you have any idea how to add a test for this ?

src/Projection/MySqlProjectionManager.php
fritz-gerneth: Not sure I follow, could you ellaborate? :)

src/Projection/PdoEventStoreProjector.php
fritz-gerneth: Not sure I follow, could you ellaborate? :)

prolic:
Should be another const in PdoEventStoreProjector, or even at interface
level.

fritz-gerneth:
Certainly can add some documentation. Any prefered location or just create a new file "Using from CLI"?

For testing - unfortunately I don't have any idea that does not require dirty hacks. The two problems I see:
- How to assert invocation of a core function (without rootkit, namespacing tricks) when doing a UT
- How spawn a new process running the projection from a test and send send a kill signal as an integration test. Technically we sure can spawn a new child process and kill it. But pcntl functions are known to cause a bit of a headache when child processes install signal handlers ([e.g. this comment](http://php.net/manual/de/function.pcntl-signal-dispatch.php#114234))
- How to test for extension loaded / not loaded 

I sure could change this to remove the dependency on `pcntl_signal_dispatch` and replace it with some generic `lifecycle callback` that can be passed to the constructor instead. I find this quite obfuscated though :)

src/Projection/PdoEventStoreProjector.php
```diff
@@ -38,6 +38,10 @@
 
 final class PdoEventStoreProjector implements Projector
 {
+    public const OPTION_PCNTL_DISPATCH = 'trigger_pcntl_dispatch';
```
prolic: I think the consts should go here: https://github.com/prooph/event-store/blob/master/src/Projection/Projector.php#L25

prolic: It's useful for all implementations, not specific to pdo

prolic:
@fritz-gerneth A unit test is not possible, I agree, but we could add an integration test. So spin up an in-memory event-store, that some test projection on it, send a kill signal and see if it stopped gracefully. 

hint:
```
$pid = exec('php test-projector.php &', $output);
posix_kill($pid, SIGUSR1);
```

src/Projection/PdoEventStoreProjector.php
fritz-gerneth: Thought about that too, they definitely belong there and are helpful for others. This would require a new release of the event-store and update the minimum version required in this package though (else the constants in the projection managers are not defined). If you prefer this I certainly can draft this up.

prolic:
Go for it!

tests/Projection/PdoEventStoreProjectorTest.php
```diff
@@ -168,4 +168,35 @@ public function it_throws_exception_when_unknown_event_store_instance_passed():
...
+        $descriptorSpec = [
+            0 => ['pipe', 'r'],
+            1 => ['pipe', 'w'],
+            2 => ['pipe', 'w'],
+        ];
+        /**
+         * Created process inherits env variables from this process.
+         * Script returns with non-standard code SIGUSR1 from the handler and -1 else
+         */
+        $projectionProcess = proc_open($command, $descriptorSpec, $pipes);
```
fritz-gerneth: This turned out to be harder than I thought and brought me down a road I certainly don't want to head to :) 

Problem with exec is that it waits for the command to finish and doesn't return the pid. We could resolve this by sending the command to the background (`php ... &`) or return the pid (`php ... & echo $!;`). But still then, it doesn't return if the output is not redirected to `/dev/null` (see comment in [documentation](http://php.net/manual/en/function.exec.php)). SO I opted for the `proc*` family instead. 

The test itself is rather simple - we return a non-standard exit code in our callback instead of the default one. Disabling the flag on the projection makes this test fail.

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -151,6 +161,7 @@ public function __construct(
         $this->persistBlockSize = $persistBlockSize;
         $this->sleep = $sleep;
         $this->status = ProjectionStatus::IDLE();
+        $this->triggerPcntlSignalDispatch = $triggerPcntlSignalDispatch && extension_loaded('pcntl');
```
prolic: if $triggerPcntlSignalDispatch is set to true but extension is missing, we should maybe throw an exception instead.

tests/Projection/PdoEventStoreReadModelProjectorTest.php
```diff
@@ -191,4 +191,39 @@ public function it_throws_exception_when_unknown_event_store_instance_passed():
             10
         );
     }
+
+    /**
+     * @test
+     */
+    public function it_dispatches_pcntl_signals_when_enabled(): void
```
prolic: We should move these tests to the AbstractReadModelProjectorTest in the event-store repo. This way the test can be reused across all event-store implementations. If you think InMemoryEventStore is not worth adding this feature, we simply skip the tests for it.

tests/Projection/isolated-read-model-projection.php
```diff
@@ -0,0 +1,71 @@
...
+use Prooph\EventStore\Pdo\MySqlEventStore;
+use Prooph\EventStore\Pdo\PersistenceStrategy\MySqlSimpleStreamStrategy;
+use Prooph\EventStore\Pdo\Projection\MySqlProjectionManager;
+use Prooph\EventStore\Pdo\Projection\PdoEventStoreProjector;
+use ProophTest\EventStore\Mock\UserCreated;
+use ProophTest\EventStore\Pdo\TestUtil;
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+$readModel = new class() implements \Prooph\EventStore\Projection\ReadModel {
```
prolic: import Prooph\EventStore\Projection\ReadModel

src/Projection/PdoEventStoreReadModelProjector.php
fritz-gerneth: I think this would limit portability of projections across systems (e.g. when a library defines a projection with this flag enabled, while the user's system doesn't have it enabled). Personally I prefer the degraded functionality approach, but it's your call.

prolic:
@codeliner your thoughts?

On Jun 30, 2017 9:41 PM, "Fritz Gerneth" <notifications@github.com> wrote:

*@fritz-gerneth* commented on this pull request.
------------------------------

In src/Projection/PdoEventStoreReadModelProjector.php
<https://github.com/prooph/pdo-event-store/pull/96#discussion_r125041499>:

> @@ -151,6 +161,7 @@ public function __construct(
         $this->persistBlockSize = $persistBlockSize;
         $this->sleep = $sleep;
         $this->status = ProjectionStatus::IDLE();
+        $this->triggerPcntlSignalDispatch =
$triggerPcntlSignalDispatch && extension_loaded('pcntl');

I think this would limit portability of projections across systems (e.g.
when a library defines a projection with this flag enabled, while the
user's system doesn't have it enabled). Personally I prefer the degraded
functionality approach, but it's your call.

—
You are receiving this because you were assigned.

Reply to this email directly, view it on GitHub
<https://github.com/prooph/pdo-event-store/pull/96#discussion_r125041499>,
or mute the thread

prolic:
Require https://github.com/prooph/event-store/releases/tag/v7.2.0 and update according to changes made there.

src/Projection/PdoEventStoreProjector.php
```diff
@@ -164,6 +170,7 @@ public function __construct(
         $this->persistBlockSize = $persistBlockSize;
         $this->sleep = $sleep;
         $this->status = ProjectionStatus::IDLE();
+        $this->triggerPcntlSignalDispatch = $triggerPcntlSignalDispatch && extension_loaded('pcntl');
```
prolic: see: https://github.com/prooph/event-store/blob/master/src/Projection/InMemoryEventStoreProjector.php#L124-L126

src/Projection/PdoEventStoreProjector.php
```diff
@@ -857,4 +866,11 @@ private function prepareStreamPositions(): void
 
         $this->streamPositions = array_merge($streamPositions, $this->streamPositions);
     }
+
+    private function triggerPcntlSignalDispatch(): void
```
prolic: Remove this private function and call it directly. It has only a single usage.

src/Projection/PdoEventStoreReadModelProjector.php
prolic: see: https://github.com/prooph/event-store/blob/master/src/Projection/InMemoryEventStoreProjector.php#L124-L126

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -807,4 +816,11 @@ private function prepareStreamPositions(): void
 
         $this->streamPositions = array_merge($streamPositions, $this->streamPositions);
     }
+
+    private function triggerPcntlSignalDispatch(): void
```
prolic: same here. we don't need it as method.

tests/Projection/isolated-read-model-projection.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
```
prolic: docheader missing

tests/Projection/isolated-projection.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
prolic: docheader missing

prolic:
thanks a lot! :+1:

Issue #98, opened 2017-07-10
[Projections] ERROR:  value too long for type character(26)


Lctrs:
While working on a new project, I noticed that my docker containers running my projections died after a random amount of time with the following error : 

`[2017-07-07 16:55:37] console.ERROR: Error thrown while running command "my_command_name". Message: "Error 22001. Maybe the projection table is not setup? Error-Info: ERROR:  value too long for type character(26)" {"error":"[object] (Prooph\\EventStore\\Exception\\RuntimeException(code: 0): Error 22001. Maybe the projection table is not setup?\nError-Info: ERROR:  value too long for type character(26) at /srv/app/vendor/prooph/pdo-event-store/src/Projection/PdoEventStoreReadModelProjector.php:727)","command":"my_command_name","message":"Error 22001. Maybe the projection table is not setup?\nError-Info: ERROR:  value too long for type character(26)"} []`

Further debugging revealed that, sometimes, the variable `$lockUntilString` in `PdoEventStoreReadModelProjector::updateLock()` holds a string that is 27 char long (for example in my case : `2017-07-07T16:55:37.1000000`).

I'm running PHP 7.1.7 in Alpine with PostgreSQL as database.

prolic:
1000000 microseconds is a second. So this should never happen. Maybe it's a PHP bug? I am not sure yet. Will do some testing.

Lctrs:
Yet it happened. For debugging, I just manually added the value of `$lockUntilString` to the error message, waited for the error to happen, and got this value. I'll do some research for a PHP bug.

Just in case, here is my full debug error message : 

[...]

PR #99, opened 2017-07-10
fix lock until string

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/98#issuecomment-314066675

The problem comes from

`$dateTime->modify('+200 ms');` is invalid, you can only add seconds, minutes, hours, and so on, but no micro- or milliseconds. This this needed to be worked around.

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -725,7 +725,8 @@ EOT;
     {
         $now = new DateTimeImmutable('now', new DateTimeZone('UTC'));
         $nowString = $now->format('Y-m-d\TH:i:s.u');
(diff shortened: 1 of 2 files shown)


src/Projection/PdoEventStoreProjector.php
```diff
@@ -758,7 +759,8 @@ private function updateLock(): void
     {
         $now = new DateTimeImmutable('now', new DateTimeZone('UTC'));
         $nowString = $now->format('Y-m-d\TH:i:s.u');
```
Lctrs: `$nowString` isn't used. It should be removed.

prolic: updated

prolic:
ping @codeliner 

please review this one as well - it needs to be merged and released asap, thanks

codeliner:
@prolic LGTM

Issue #97, opened 2017-06-30
Replace real_stream_name LIKE ? with category = ? IN projectors and queries


simon-ondrej:
The default table structure contains an index on category attribute ([MySql](https://github.com/prooph/pdo-event-store/blob/master/scripts/mysql/01_event_streams_table.sql#L9), [Postgres](https://github.com/prooph/pdo-event-store/blob/master/scripts/postgres/01_event_streams_table.sql#L10), [MariaDb](https://github.com/prooph/pdo-event-store/blob/master/scripts/mariadb/01_event_streams_table.sql#L10)) but when projections or queries `fromCategory` or `fromCategories` are made, projection and query classes do not utilize this index and do a [`LIKE` comparison on the `real_stream_name`](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreQuery.php#L391) ([Projector](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreProjector.php#L855)) attribute instead - which is slower.

The proposal is to use the indexed field instead, considering the values in `real_stream_name` are generated as `<category>-<aggregate_id>` anyway this should not be a problem.

Additional note: It might also be worth converting the where to an `IN` expression rather than a combination of `OR`s.

prolic:
:+1:

PR #100, opened 2017-07-10
optimize fromCategories event store queries

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/97

/cc @simon-ondrej


codeliner:
@prolic LGTM

Issue #103, opened 2017-08-06
Connection preprty never used in PdoStreamIterator


sandrokeil:
The connection property is never used in [PdoStreamIterator](https://github.com/prooph/pdo-event-store/blob/master/src/PdoStreamIterator.php#L91), so we can remove it.

PR #104, opened 2017-08-07
remove connection property from stream iterator

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/103

/cc @sandrokeil


sandrokeil:
@prolic LGTM but do we have a BC break here? I'm not sure if `PdoStreamIterator` is used in the wild.

prolic:
It's an implementation details of the event store. You can't extend it in
any way. That's why it's no BC to me.

PR #105, opened 2017-08-21
PdoEventStore Interface

oqq:
This PR provides a PdoEventStore interface (without any additional method for now).

Goal is to remove this instance chaining test expression, so that it would not be necessary to change that expression in future implementations.
Also it would be possible to use own pro event store implementations by users.

Another Proposal (not implemented yet):
The PdoEventStore interface could provide a `getConnection` method to ensure the same connection is used for projection. (Not sure if that would be intentional)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,6 +65,11 @@
         "cs-fix": "php-cs-fixer fix -v --diff",
         "test-postgres": "DB_HOST=postgres vendor/bin/phpunit -c phpunit.xml.postgres",
         "test-mariadb": "DB_HOST=mariadb vendor/bin/phpunit -c phpunit.xml.mariadb",
-        "test-mysql": "DB_HOST=mysql vendor/bin/phpunit -c phpunit.xml.mysql"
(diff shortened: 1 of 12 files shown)


prolic:
@codeliner what do you think about this change?

prolic:
My thoughs on this quote `The PdoEventStore interface could provide a getConnection method to ensure the same connection is used for projection`. What I dislike about this, is that it exposes the db connection to the outside world, which I don't like. The interface should be clean, you should only be able to do something with the object, that you are intended to do.

codeliner:
LGTM - even if PdoEventStore is just a marker interface it is a good addition. All PDO event stores are in the pdo package so a common PdoEventStore interface makes sense.

I'm with @prolic regarding connection. It is a good thing that the new interface allows us to add some pdo specific methods if needed. But we should be careful with this.

oqq:
I am fine with your arguments against the `getConnection` method. 👍

@prolic 
Is it intended to use the same connection for projection and event store?

prolic:
You should use same connection, but it's okay to have a different one, as
long as you're connected to the same db.

prolic:
thanks a lot

PR #107, opened 2017-09-18
Fix mariadb

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/106

/cc @Danielss89

@sandrokeil @codeliner better ideas? mariadb seems pretty stupid here.


Danielss89:
Works for me!

prolic:
I know it works, but I don't like the addslashes done in php. Is there no
better solution?

prolic:
Travis failed on mariadb because `ERROR 1006 (HY000) at line 1: Can't create database 'event_store_tests' (errno: 2 "No such file or directory")` - some travis problem again, they are changing their system way to often to break tests. Let's ignore this for now.

Danielss89:
Hm well, i had an idea to just do `json_encode` on the value, but that adds a starting and ending quote too, so it would have to be `trim(json_encode($v), '"');` which is also a bit ugly.

prolic:
MariaDB is shitty :)

sandrokeil:
@prolic Have some issues local with Docker and MariaDB 10.2 
> Tests: 190, Assertions: 1058, Errors: 8, Failures: 2.

prolic:
@sandrokeil travis build is fixed and maria db tests green again.

sandrokeil:
@prolic Tests are green now :+1: Have you tried to work with [JSON_OBJECT](https://dev.mysql.com/doc/refman/5.7/en/json.html#json-values)? MariaDB has same functions but I don't know if this works.

prolic:
@sandrokeil `JSON_OBJECT` doesn't help, as we need key + value for it to work.

PR #108, opened 2017-10-05
Drop stream table only if it exists

sergeyfedotov:
If an error occurs while creating a stream table, the stream table dropping produces a new error because the table doesn't exists. In this case `removeStreamFromStreamsTable` is not called and a corresponding record from the streams table is not deleted.

### DIFF
--- a/tests/MariaDbEventStoreTest.php
+++ b/tests/MariaDbEventStoreTest.php
@@ -20,6 +20,7 @@ use Prooph\EventStore\Metadata\MetadataMatcher;
 use Prooph\EventStore\Metadata\Operator;
 use Prooph\EventStore\Pdo\Exception\RuntimeException;
 use Prooph\EventStore\Pdo\MariaDbEventStore;
+use Prooph\EventStore\Pdo\PersistenceStrategy;
 use Prooph\EventStore\Pdo\PersistenceStrategy\MariaDbAggregateStreamStrategy;
 use Prooph\EventStore\Pdo\PersistenceStrategy\MariaDbSingleStreamStrategy;
 use Prooph\EventStore\Stream;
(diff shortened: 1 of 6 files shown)


prolic:
Can you provide a test case that is failing without this change?

prolic:
ping @sergeyfedotov

tests/MariaDbEventStoreTest.php
```diff
@@ -173,4 +174,28 @@ public function it_ignores_transaction_handling_if_flag_is_enabled(): void
 
         $eventStore->appendTo(new StreamName('Prooph\Model\User'), new ArrayIterator([$streamEvent]));
     }
+
+    /**
+     * @test
+     */
+    public function it_removes_stream_if_stream_table_doesnt_exist(): void
```
prolic: we should move this test to AbstractPdoEventStoreTest

sergeyfedotov: I've written this test case for MariaDB as an example. The query that I used to raise an exception isn't portable between different databases. If you think it should be moved to the parent class, it might be better to write a more common query. Another solution is to write a query with invalid syntax. What do you think?

sergeyfedotov: It's also creates a database-specific event store object. I think it's better to write database-specific test cases.

sergeyfedotov:
I've provided test cases for all event stores.

prolic:
thanks @sergeyfedotov - I will check this again tomorrow. How did you encounter this issue? The test-case covers how to reproduce it with mocks, but I am not sure how to get to this in a production environment, can you explain that to me?

sergeyfedotov:
In my case, it was using mariadb 10.1 while minimum required version is 10.2.3. I had to manually clear streams table after upgrade.

prolic:
thanks a lot

PR #111, opened 2017-10-27
Catch Throwable in transactional()

jiripudil:
`transactional()` should rollback on any \Throwable, not just exceptions.


prolic:
good catch, thanks

PR #112, opened 2017-10-27
PdoStreamIterator should check json_last_error after json_decode

denniskoenig:
I am not sure about a unit test and if it is ok to use an own exception. Is this a correct style for prooph?

### DIFF
--- a/src/Exception/JsonException.php
+++ b/src/Exception/JsonException.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * This file is part of the prooph/pdo-event-store.
+ * (c) 2016-2017 prooph software GmbH <contact@prooph.de>
+ * (c) 2016-2017 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo\Exception;
+
+use Prooph\EventStore\Exception\RuntimeException as EventStoreRuntimeException;
+
+class JsonException extends EventStoreRuntimeException implements PdoEventStoreException
+{
+    public static function whileDecode(string $msg, int $code, string $json): JsonException
+    {
+        return new self(
+            sprintf(
+                "Error while decoding JSON. \nMessage: %s \nJSON: %s",
+                $msg,
+                $json
+            ),
+            $code
+        );
+    }
+}
--- a/src/PdoStreamIterator.php
+++ b/src/PdoStreamIterator.php
@@ -20,6 +20,7 @@ use PDOException;
 use PDOStatement;
 use Prooph\Common\Messaging\Message;
 use Prooph\Common\Messaging\MessageFactory;
+use Prooph\EventStore\Pdo\Exception\JsonException;
 use Prooph\EventStore\Pdo\Exception\RuntimeException;
 
 final class PdoStreamIterator implements Iterator
@@ -120,10 +121,19 @@ final class PdoStreamIterator implements Iterator
             $metadata['_position'] = $this->currentItem->no;
         }
 
+        $payload = json_decode($this->currentItem->payload, true);
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw JsonException::whileDecode(
+                json_last_error_msg(),
+                json_last_error(),
+                $this->currentItem->payload
+            );
+        }
+
         return $this->messageFactory->createMessageFromArray($this->currentItem->event_name, [
             'uuid' => $this->currentItem->event_id,
             'created_at' => $createdAt,
-            'payload' => json_decode($this->currentItem->payload, true),
+            'payload' => $payload,
             'metadata' => $metadata,
         ]);
     }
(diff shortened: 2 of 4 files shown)


src/Exception/JsonException.php
```diff
@@ -0,0 +1,30 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo\Exception;
+
+use Prooph\EventStore\Exception\RuntimeException as EventStoreRuntimeException;
+
+class JsonException extends EventStoreRuntimeException implements PdoEventStoreException
```
prolic: @codeliner your thoughts on this?

src/PdoStreamIterator.php
```diff
@@ -120,10 +121,19 @@ public function current(): ?Message
             $metadata['_position'] = $this->currentItem->no;
         }
 
+        $payload = json_decode($this->currentItem->payload, true);
+        if (json_last_error() !== JSON_ERROR_NONE) {
```
prolic: unit test should be simple, just insert some garbage to the database in a test-case

src/Exception/JsonException.php
codeliner: I like it especially the method `whileDecode`

prolic:
@denniskoenig Can you provide a unit test, please?

denniskoenig:
I am not sure how.
- Would the test go into MySqlEventStoreTest.php ?
- I think the charset of the connection must be changed to create malformed characters. What is the best way to do in the test (via SET NAMES)?

prolic:
@denniskoenig I think AbstractPdoEventStoreTest would be a nice fit. Actually you don't need to change the charset. You can simply insert invalid json manually (f.e. invalid characters, syntax error, whatever). It should be sufficient, to do $this->eventStore->load(...) and check if the expected exception is received.

denniskoenig:
Sorry took a while, because i got some other stuff to do. Unfortunately I can not trigger the exception. Can anyone help?

tests/AbstractPdoEventStoreTest.php
```diff
@@ -486,6 +487,31 @@ public function it_doesnt_double_escape_metadata(): void
...
+
+        $event = UserCreated::with(['name' => ['John', 'Jane']], 1);
+        $event = $event->withAddedMetadata('key', 'value');
+
+        $streamName = new StreamName('Prooph\Model\User');
+        $stream = new Stream($streamName, new ArrayIterator([$event]), ['some' => ['metadata', 'as', 'well']]);
+
+        $this->eventStore->create($stream);
+
+        // TODO: Make json somehow invalid to trigger exception.
```
prolic: Grab the db connection and update the row with a raw query. This for example would be invalid: `{{`

denniskoenig: MySQL prevents inserting invalid json

prolic:
Really? MySQL is clever :-) then you probably never get invalid json back

tests/AbstractPdoEventStoreTest.php
denniskoenig: only if there is a fuck up with the encoding...

prolic:
Then you could create a test case with wrong encoding on db connection and
pass that to the event store instance

tests/AbstractPdoEventStoreTest.php
prolic: maybe we should create a custom connection with wrong encoding to get the test failing?

prolic:
@denniskoenig I really want to get this merged and released asap, can you finish the tests this weekend?

tests/AbstractPdoEventStoreTest.php
denniskoenig: I have no time on the weekend, but I'll finish it on monday.

prolic:
Thanks

tests/AbstractPdoEventStoreTest.php
```diff
@@ -486,6 +486,34 @@ public function it_doesnt_double_escape_metadata(): void
...
+        // Trigger an error when using an umlaut in the payload.
+        $this->connection->query('SET NAMES latin1');
+
+        $metadataMatcher = new MetadataMatcher();
+        $metadataMatcher = $metadataMatcher->withMetadataMatch(
+            'event_id', Operator::EQUALS(), $event->uuid()->toString(), FieldType::MESSAGE_PROPERTY()
+        );
+
+        $streamEvents = $this->eventStore->load($streamName, 1, null, $metadataMatcher);
+        $streamEvents->current(); // Trigger PdoStreamIterator.
```
denniskoenig: I am not sure if i used the loading correct. Is there a better way to trigger the PdoStreamIterator?

prolic: that's good, problem is those tests for mariadb, and postgres are failing.

denniskoenig: damn. any idea to make the test more generic or move the test to MySqlEventStoreTest and write tests especially for mariadb etc?

prolic: I don't know, I'm also good with some specific tests for each db type, at least we have it covered then.

prolic:
@denniskoenig we have a merge conflict on `tests/MySqlEventStoreTest.php`. Any chance you get a test for postgres and mariadb running as well?

denniskoenig:
I will try to create a test for mariadb this evening. sorry that it is going slowly, at the moment i have a lot other things to do.

tests/AbstractPdoEventStoreTest.php
denniskoenig: I have to play around a bit with mariadb and postgres (i am not working with them) but this will take some time. if anyone knows how to create an error while json_decode with mariadb/postgres it would speed up the things.
The next two weeks i'm still busy and can not promise i will find the time to get this done.
If you know mariadb/postgres it's probably a trifle to get the tests done.

PR #115, opened 2017-11-21
Fixes position in projections

Adapik:


prolic:
I need to take a deeper look into it, as this might affect other places as well.

Adapik:
@prolic Take a deep look, please. It may affect some places in client code when someone loads events and picks them by key `$event = $events[0]` as long as keys are not 0,1,2.. etc. Someone could rely on that. In package itself (and the other packages) I haven't found such usages. Also If you'll see another solution of the original problem please describe it.

prolic:
@Adapik Sorry I'm really busy these days, will try to solve this at the weekend.

codeliner:
LGTM

In the past we said that no event should be missing in the stream so that position worked. I think we don't break the old expectation because as long as no event is missing the position of the event is the same as the array index.
The advantage now is that an event can be missing but stream position still works fine. So :+1: for the change.

PR #117, opened 2017-12-05
[WIP] Refactored error handling

enumag:
I want to add previous exceptions everywhere, add some static factories to reduce code duplication etc. It's far from ready. Feedback is appreciated though.

### DIFF
--- a/src/PostgresEventStore.php
+++ b/src/PostgresEventStore.php
@@ -81,6 +81,9 @@ final class PostgresEventStore implements PdoEventStore, TransactionalEventStore
 
         Assertion::min($loadBatchSize, 1);
 
+        $this->setAttribute(PDO::ATTR_STATEMENT_CLASS, ['Doctrine\DBAL\Driver\PDOStatement', []]);
+        $this->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+
         $this->messageFactory = $messageFactory;
         $this->connection = $connection;
         $this->persistenceStrategy = $persistenceStrategy;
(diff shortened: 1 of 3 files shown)


src/PostgresEventStore.php
```diff
@@ -81,6 +81,9 @@ public function __construct(
 
         Assertion::min($loadBatchSize, 1);
 
+        $this->setAttribute(PDO::ATTR_STATEMENT_CLASS, ['Doctrine\DBAL\Driver\PDOStatement', []]);
```
enumag: Inspired from Doctrine/DBAL.

codeliner: but Doctrine/DBAL is not a dependency, so why do you set it as statement class?

src/PostgresEventStore.php
```diff
@@ -81,6 +81,9 @@ public function __construct(
...
+        $this->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
```
prolic: I don't want to force the user to a specific usage here.

src/PostgresEventStore.php
enumag: Sorry that was a mistake. I'll get rid of that.

src/PostgresEventStore.php
enumag: Can you explain? The PDO is not intended to be used outside of the EventStore, right? So it should be ok to use `PDO::ERRMODE_EXCEPTION`. And honestly noone should ever use PDO without this - the need to check errors manually leads to critical mistakes like #118 which is what I want to prevent.

prolic: The problem is, that someone with access to the PDO instance, can change this attribute afterwards and the event-store stops working suddenly. The bug has been fixed and I don't expect to get another one popping up because of this. I think we leave it as is.

enumag: In that case we should create the PDO in constructor instead of require it. I actually do expect furhter problems caused by this. Hopefully not as severe but the exceptions thrown by the EventStore are quite inconsistent now and there is a lot of duplicated code. It does need refactoring.

prolic: code duplication is not an issue. what do you mean with inconsistency in exceptions? can you show some examples?

prolic: changing the constructor is a bc break and i will not do this now.

PR #118, opened 2017-12-05
Fix silently ignored errors in PostgresEventStore

enumag:
This hotfix should be merged an released ASAP. Better fix will be later in https://github.com/prooph/pdo-event-store/pull/117.

### DIFF
--- a/src/PostgresEventStore.php
+++ b/src/PostgresEventStore.php
@@ -219,6 +219,10 @@ EOT;
         if (in_array($statement->errorCode(), ['23000', '23505'], true)) {
             throw new ConcurrencyException();
         }
+        
+        if ($statement->errorCode() !== '00000') {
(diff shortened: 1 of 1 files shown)


sandrokeil:
Looks like this check was missed, are other files also affected?

enumag:
I didn't have time to check them thoroughly but from a quick glance it didn't seem to. Still I'll be changing the error handling to exceptions in #117 which will fix any other potential instances of this bug as a side-effect.

enumag:
@prolic The test failure is not caused by this PR.

codeliner:
LGTM for a hotfix, @prolic or @oqq second go?

error is caused by mariadb, probably because of the new mariadb version that causes trouble on our end.

oqq:
jep.. looks fine

Issue #122, opened 2017-12-23
Update schemas


prolic:
See here: https://github.com/prooph/event-store-benchmarks/issues/6

PR #124, opened 2018-01-19
upgrade table schemas

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/122

/cc @oqq @codeliner


prolic:
@hvanoch @basz wanna check this as wel?

fritz-gerneth:
To clarify my stance on this:
- Changing the datatypes in the DB is a BC, there might be some code depending on this, for whatever reason (for either column)
- Depending on your setup & team, a DB migration requires a bit more careful planning & evaluation than just a minor update / patch update.
- My Setup break after a composer update without changing any major version, I first have to figure out what package did wha. Composer does not provide any update instructions so first I'm blind which package broke at all.

'Sum:
- If MariaDB projections are not usable at all due to this, I'd consider this ok to fix with a BC break for mariadb (as they were not operational in the first place there was nothing to break)
- If MariaDB projections were functional but only could store a small state I'd consider this a BC as those using it will have their setup broken at first
- Postgress, this is a BC for any user too, for the sole purpose of optimization, not to fix any ciritial bug. I'd be not amussed if I had to perform a migration after a minor/patch level update. This really should wait for a major release.

codeliner:
I'm with @fritz-gerneth . A simple composer update can pull in the breaking changes and I'm not aware on what is going on. Why not simply release a pdo-event-store v2 ? And fixing only the mariaDB issue with a 1.6.1 release?

When we communicate major versions we only use the versions of the main repos in that case event store v7 and that version would not change. You would simply use event-store v7 with pdo-event-store v2 if you are ready to migrate your DB

prolic:
Ok here is the thing, nothing breaks! Even if you don't upgrade the database, everything still works. 

MySQL - no changes at all
Postgres - Migrating will optimize db tables, not migrating and still running newer version and everything still works. Only difference is using aggregate stream strategy: new tables will have new schema, where older tables have old schema. But nothing breaks, everything works.
MariaDB - this is broken either way, so I think a minor release is okay, as MariaDB will break currently with current projection table, at least very easy when using aggregate stream strategy. Hence the MariaDB update is really required.

prolic:
Additional note: I'm okay with dropping the change for postgres, but I would prefer having MariaDB change done for 1.7 release.

codeliner:
ah ok @prolic thanks for clarification. Thought the schema changes would have more effect. Let's add one more sentence to the docs to point that out. At the moment it reads like: "no big changes, but you have to perform those schema migrations in the background". Just add something like: "For Postrgres you can do the migration later. Current schema will continue to work with v1.7 but the schema migration is recommended"

fritz-gerneth:
Was under the same impression as @codeliner . If the migration is optional (though recommended) its not a (major) BC.

PR #126, opened 2018-01-23
Always update the projections position

### LINKED ISSUE #125: Always update the projections position
Now the projection position is only updated when it actually has an event.
That means when stopping the projection and running it again, all events will be fetched from previous position to "now".

The real reason I want this to be updated is so I can update the read model synchronised, and not async with the write model.
I created a plugin that when ever a write happens (event store "appendTo" or "create" event) waits and poll for the position to be the same as the key from last event created. That way I know that the projection has the current state applied of these newly created events.

As you see, if the position is never updated for projections where the event was never used for it hangs.

hvanoch:
Fixes #125

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -756,12 +756,18 @@ EOT;
         $lockUntilString = $this->createLockUntilString($now);
 
         $sql = <<<EOT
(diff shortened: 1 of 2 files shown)


prolic:
@hvanoch Can you update the Projector (not ReadModel-Projector) as well?

hvanoch:
@prolic done.

prolic:
Thank you very much!

PR #130, opened 2018-02-05
Quote tablenames

basz:
refactored a bit. but at this point tests still pass.

### DIFF
--- a/tests/Assets/PersistenceStrategy/MySqlSimpleStreamStrategy.php
+++ b/tests/Assets/PersistenceStrategy/MySqlSimpleStreamStrategy.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * This file is part of the prooph/pdo-event-store.
+ * (c) 2016-2017 prooph software GmbH <contact@prooph.de>
+ * (c) 2016-2017 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace ProophTest\EventStore\Pdo\Assets\PersistenceStrategy;
+
+use Iterator;
+use Prooph\EventStore\Pdo\PersistenceStrategy;
+use Prooph\EventStore\StreamName;
+
+final class MySqlSimpleStreamStrategy implements PersistenceStrategy
+{
+    /**
+     * @param string $tableName
+     * @return string[]
+     */
+    public function createSchema(string $tableName): array
+    {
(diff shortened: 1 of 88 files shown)


tests/Assets/PersistenceStrategy/MySqlSimpleStreamStrategy.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * This file is part of the prooph/pdo-event-store.
+ * (c) 2016-2017 prooph software GmbH <contact@prooph.de>
```
prolic: 2018

tests/Assets/PersistenceStrategy/MySqlSingleStreamStrategy.php
```diff
@@ -0,0 +1,84 @@
...
+declare(strict_types=1);
+
+namespace ProophTest\EventStore\Pdo\Assets\PersistenceStrategy;
+
+use Iterator;
+use Prooph\EventStore\Pdo\HasQueryHint;
+use Prooph\EventStore\Pdo\PersistenceStrategy;
+use Prooph\EventStore\StreamName;
+
+final class MySqlSingleStreamStrategy implements PersistenceStrategy, HasQueryHint
```
prolic: Maybe rename to CustomMySqlSingleStreamStrategy

tests/MySqlCustomStrategiesEventStoreTest.php
```diff
@@ -0,0 +1,114 @@
+<?php
+/**
+ * This file is part of the prooph/pdo-event-store.
+ * (c) 2016-2017 prooph software GmbH <contact@prooph.de>
```
prolic: 2018

prolic:
Looks good, some notes:
1) Docheader tests failed, that's why travis is red.
2) Please also add custom strategies for MariaDB and Postgres, so we can make sure they are running good as well.

basz:
I dunno why that one test fails. run in docker fine. looks like it couldn't initialise something proper looking at some error at the top.

secondly: I feel we should also apply the same pattern to the projection tests. not hitting the 41char limit currently.

thirdly an idea: lets integrate the event stream and projection table creation into the strategies as well?
once we hit `($statement->errorCode() !== '00000')` could try to run it, according to the choosen strategy? Just interested in your thoughts. should be separate PR and it a BC i think.

prolic:
Seems like travis was drunk. I restarted the failed build and now it's green again.

Wanna submit another PR for projections and readmodelprojections?

The event stream and projection table has to be outside the scope of persistence strategies, so you can use different strategies parallel, while still having the projectors running.

PR #131, opened 2018-02-06
Event name propagation

basz:
I think I have hit a bug that should be fixed before I can continue with quoting of table names in projections.

this PR only demonstrates it, checkout and run with;
```
docker-compose -f docker-compose-tests.yml run composer run-script test-postgres --timeout 0 tests/Projection/PostgresEventStoreReadModelProjectorTest.php; docker stop proophpdoeventstore_postgres_1 && docker rm proophpdoeventstore_postgres_1
```

something is still trying to insert into `event_streams`

### DIFF
--- a/scripts/postgres/01_event_streams_table.sql
+++ b/scripts/postgres/01_event_streams_table.sql
@@ -1,4 +1,4 @@
-CREATE TABLE event_streams (
+CREATE TABLE e_stream (
   no BIGSERIAL,
   real_stream_name VARCHAR(150) NOT NULL,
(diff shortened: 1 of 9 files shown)


prolic:
@basz in PostgresCustomStrategiesEventStoreTest you call `TestUtil::initDefaultDatabaseTables($this->connection);` - that's why

basz:
no cause I changed that one too... ;-). But please - stop reviewing - I might have found the problem...

prolic:
okay :)

PR #132, opened 2018-02-08
Quote tablenames projections

basz:
simply changes tables names of events_streams and projections, adds tests to use them. should still pass


basz:
ok i got it now. turns out that open_proc does not pass empty env vars (password) so the isset check inside the isolated tests would fail. No clue why we didn't catch this earlier and why only in err mode 2.

a small pita

basz:
ping

src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -858,6 +873,17 @@ private function createLockUntilString(DateTimeImmutable $from): string
...
-        return $from->modify('+' . $secs .' seconds')->format('Y-m-d\TH:i:s') . '.' . $resultMicros;
+        return $from->modify('+' . $secs . ' seconds')->format('Y-m-d\TH:i:s') . '.' . $resultMicros;
+    }
+
+    private function quoteTableName(string $tableName): string
+    {
+        switch ($this->vendor) {
+            case 'pgsql':
+                return '"'.$tableName.'"';
+                break;
```
prolic: break is not needed here, because we return already

tests/TestUtil.php
```diff
@@ -110,20 +115,42 @@ public static function initDefaultDatabaseTables(PDO $connection): void
...
+    public static function tearDownDatabase(): void
+    {
+        $connection = self::getConnection();
+        $vendor = self::getDatabaseVendor();
+        switch ($vendor) {
+            case 'postgres':
+                $statement = $connection->prepare('SELECT table_name FROM information_schema.tables WHERE table_schema = \'public\';');
+                break;
+            default:
+                $statement = $connection->prepare('SHOW TABLES');
```
prolic: break missing

tests/TestUtil.php
```diff
@@ -110,20 +115,42 @@ public static function initDefaultDatabaseTables(PDO $connection): void
...
+        $statement->execute();
+        $tables = $statement->fetchAll(PDO::FETCH_COLUMN);
+
+        foreach ($tables as $table) {
+            switch ($vendor) {
+                case 'postgres':
+                    $connection->exec(sprintf('DROP TABLE "%s";', $table));
+                    break;
+                default:
+                    $connection->exec(sprintf('DROP TABLE `%s`;', $table));
```
prolic: break missing

tests/TestUtil.php
```diff
@@ -134,7 +161,7 @@ private static function hasRequiredConnectionParams(): bool
             $env['DB'],
             $env['DB_DRIVER'],
             $env['DB_USERNAME'],
-            $env['DB_PASSWORD'],
+            // $env['DB_PASSWORD'], // an empty string, which does not survive open_proc
```
prolic: remove the comment, just remove the line

prolic:
👍

PR #134, opened 2018-03-12
Remove useless code

enumag:
While debugging https://github.com/proophsoftware/es-emergency-call/issues/5 I also notices there are 2 SQL queries executed every time you call `EventStore::load()`. The first query checks if the stream exists in the database and only second query selects the events.

I believe the first command is useless. In case the table doesn't exist the second command will fail and it will result in the same exception (`StreamNotFound`). The only difference is that there is no check if the stream is registered in the event streams table which doesn't seem too important.

This doesn't make any huge difference for me as my performance issue was far more serious but still halving the amount of SQL queries is nice.

### DIFF
--- a/src/PostgresEventStore.php
+++ b/src/PostgresEventStore.php
@@ -233,13 +233,6 @@ EOT;
     ): Iterator {
         $tableName = $this->persistenceStrategy->generateTableName($streamName);
 
-        $query = "SELECT stream_name FROM \"$this->eventStreamsTable\" WHERE stream_name = ?";
-        $statement = $this->connection->prepare($query);
-        $statement->execute([$tableName]);
-
-        if ($statement->rowCount() === 0) {
-            throw StreamNotFound::with($streamName);
-        }
         [$where, $values] = $this->createWhereClause($metadataMatcher);
         $where[] = 'no >= :fromNumber';
 


prolic:
@enumag As you can see the tests fail with the code removed. It was places there intentionally.

enumag:
Well I don't have time to create such benchmark obviously. And you don't actually need it either, it's simple math.

When you have an event-store projection to use as your reading database you usually need to somehow project all events. So the projector can look simply like this:

```php
		$projection = $this->projectionManager->createReadModelProjection(
			self::PROJECTION_NAME,
			$mainReadModel
		);

		$projection
			->fromAll()
			->whenAny(
				function ($state, Message $event) use ($mainReadModel): void {
					$mainReadModel->stack('onEventUpdate', $event);
				}
			)
			->run($keepRunning);
```

When `$keepRunning` is true the run method goes through all streams over and over, checking them if there are some new events in each stream. Each check is an SQL query ([or two for Postgres](https://github.com/prooph/pdo-event-store/pull/134)) or some equivalent of a different driver is used. Well let's say each check takes 1/10 ms = 100 μs.

When you have 1000 aggregates, you have 1000 streams so it takes 1 second to check all of them.

Then your database grows to 1,000,000 aggregates so it takes 16.6 minutes to check all the streams. Which means some of your events can take up to 16 minutes to be projected (depending on where in the loop the projector is when the event is added to ES). That's already quite slow.

Then your database grows again to 100,000,000 aggregates and your projection time gets to ~27 hours. And that is not even counting the time spent to actually project each event.

If this doesn't convince you then I rest my case and you can close this issue.

Either way I thank you for your valueable time.

PR #136, opened 2018-03-12
update docs

prolic:


enumag:
If I understand the standard projections correctly then they propagate an event from one stream to another so you can then pull all the events you need from that single stream.

If I get the events I need to project from a stream created by CategoryStreamProjectionRunner or AllStreamProjectionRunner it only means I've moved the original problem from from the real projections to these that create the aggregated streams.

So this however doesn't solve anything at all and OneStreamPerAggregateStrategy is of course still completely useless for production environment.

prolic:
@enumag No, it does make a huge difference, when you have multiple projections running at the same time. The work is only done once, and not every projection has to query all streams again himself.

Also: One-Stream-Per-Aggregate is not useless in production environment, it only becomes a bottleneck when using event-store projection (which you are not required to do!).

There will be a way more performant solution in the future, but this will take time to develop, you can't expect this to happen soon (unless you pay for the development).

enumag:
> No, it does make a huge difference, when you have multiple projections running at the same time. The work is only done once, and not every projection has to query all streams again himself.

Yeah you're right about that. I only have one projection though so it doesn't make a difference to me.

> Also: One-Stream-Per-Aggregate is not useless in production environment, it only becomes a bottleneck when using event-store projection (which you are not required to do!).

In any realworld application you do need a projection so this comment is completely irrelevant.

> There will be a way more performant solution in the future, but this will take time to develop, you can't expect this to happen soon (unless you pay for the development).

I don't expect a better solution to be available soon. I'm happy to use stream-per-aggregate-type until it's available. I just think that this problem should be mentioned in the documentation. The documentation currently recommends `AggregateStreamStrategy` as the best solution but it is in fact a newbie trap. Until a better solution is developed `SingleStreamStrategy` should be recommended instead.

If you disagree that projection is needed for any realworld usage then add a note that `AggregateStreamStrategy` is in fact better for the simple case when you don't have projections.

prolic:
@enumag You can create projections without the event-store-projections package. The event-bus can publish messages to the projector directly or to amqp. So there is no reason to believe, that event-store-projections are the only way to go and therefore One-Stream-Per-Aggregate is not suitable. It might be not suitable for you, but that's another story.

As you can see, I updated the docs already and added that note.

enumag:
I don't think the note is correct...

```
For event-store projections the aggregate stream strategy is not that important anymore, consider using CategoryStreamProjectionRunner...
```

As I've pointed out above doesn't really solve the bottle neck. Yes it helps when you have multiple projections but doesn't solve the main problem.

In my opinion the note should say this: 

```
The aggregate stream strategy is not suitable for event-store projections. In this case SingleStreamStrategy is recommended instead.
```

prolic:
I would need a benchmark to show that point first, before I make such an
announcement.

Sascha-Oliver Prolic

2018-03-12 23:09 GMT+08:00 Jáchym Toušek <notifications@github.com>:

enumag:
Well I don't have time to create such benchmark obviously. And you don't actually need it either, it's simple math.

When you have an event-store projection to use as your reading database you usually need to somehow project all events. So the projector can look simply like this:

```php
		$projection = $this->projectionManager->createReadModelProjection(
			self::PROJECTION_NAME,
			$mainReadModel
		);

		$projection
			->fromAll()
			->whenAny(
				function ($state, Message $event) use ($mainReadModel): void {
					$mainReadModel->stack('onEventUpdate', $event);
				}
			)
			->run($keepRunning);
```

When `$keepRunning` is true the run method goes through all streams over and over, checking them if there are some new events in each stream. Each check is an SQL query ([or two for Postgres](https://github.com/prooph/pdo-event-store/pull/134)) or some equivalent of a different driver is used. Well let's say each check takes 1/10 ms = 100 μs.

When you have 1000 aggregates, you have 1000 streams so it takes 1 second to check all of them.

Then your database grows to 1,000,000 aggregates so it takes 16.6 minutes to check all the streams. Which means some of your events can take up to 16 minutes to be projected (depending on where in the loop the projector is when the event is added to ES). That's already quite slow.

Then your database grows again to 100,000,000 aggregates and your projection time gets to ~27 hours. And that is not even counting the time spent to actually project each event.

If this doesn't convince you then I rest my case and you can close this issue.

Either way I thank you for your valueable time.

prolic:
@enumag I updated the docs again, now better?

docs/variants.md
```diff
@@ -65,11 +65,20 @@ This stream strategy should be used together with event-sourcing, if you use one
...
 
-This stream strategy is the most performant of all, but it will create a lot of database tables, which is something not
+This stream strategy is the most performant of all (with downsides, see notes), but it will create a lot of database tables, which is something not
 everyone likes (especially DB admins).
 
 All needed database tables will be created automatically for you.
 
+Note: For event-store projections the aggregate stream strategy is not that performant anymore, consider using [CategoryStreamProjectionRunner](https://github.com/prooph/standard-projections/blob/master/src/CategoryStreamProjectionRunner.php) from the [standard-projections]((https://github.com/prooph/standard-projections) repository.
+But even than, the projections would be slow, because the projector would need to check all streams for new events, let's take this example:
+- Each check is an SQL query, well let's say each check takes 1ms.
```
enumag: You increased this estimate 10 times so the numbers below don't match anymore.

I'm not sure we should put such numbers to documentation either. As you said, we don't have a benchmark, it's just very rough estimate (in reality the times would probably be higher).

How about this?

```
But even than, the projections would be slow, because the projector needs to check all the streams one-by-one for any new events. Because of this speed of finding and projecting any new events depends on the number of streams which means it would rapidly decrease as you add more data to your event store.

You could however drastically improve the projections, if you would add a category stream projection as event-store-plugin. (This doesn't exist, yet)
```

prolic: accepted :)

enumag:
@prolic LGTM, thanks! :-)

Issue #138, opened 2018-03-14
wrong ProjectionNotFound exception on resetProjection


oqq:
If I run `MySqlProjectionManager::resetProjection` on a projection which has the status 'resetting', this method would throw a ProjectionNotFound exception.

This occurs probably since `$statement->rowCount()` would also return 0 if no row was changed at all.

Maybe also true for Postgres and MariaDB and of course other methods like `stop` and `delete`.

prolic:
confirmed bug in MySQL and MariaDB, no problem with Postgres

prolic:
working on a fix...

PR #140, opened 2018-03-25
fix multiple calls to reset/stop/delete projection

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/138
see: https://github.com/prooph/event-store/pull/329 (this has to be merged an released first, so we have tests included)

/cc @oqq

### DIFF
--- a/src/Projection/MariaDbProjectionManager.php
+++ b/src/Projection/MariaDbProjectionManager.php
@@ -138,7 +138,23 @@ EOT;
         }
 
         if (0 === $statement->rowCount()) {
-            throw ProjectionNotFound::withName($name);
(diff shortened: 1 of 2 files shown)


prolic:
Nice hint, let me try that....

prolic:
I tried this:
```
        if (0 === $statement->rowCount() && 0 === $statement->getAttribute(PDO::MYSQL_ATTR_FOUND_ROWS)) {
            throw ProjectionNotFound::withName($name);
        }
```

result:

`PDOStatement::getAttribute(): SQLSTATE[IM001]: Driver does not support this function: This driver doesn't support getting attributes`

PR #149, opened 2018-05-02
Fix MariaDB indexed queries

### LINKED ISSUE #147: MariaDB does not use INDEX
When querying for aggregates this query is being used:
```
SELECT * FROM `{MY_STREAM}` USE INDEX(ix_query_aggregate)
WHERE json_value(metadata, '$._aggregate_type') = '{MY_TYPE}'  AND json_value(metadata, '$._aggregate_id') = '{MY_AGGR}'  AND `no` >= 1
ORDER BY `no` ASC
LIMIT 10000
```
Unlike MySQL, MariaDB doesn't use indexes in these case, therefor the actual columns (which already exists) need to be used. 
This causes the query to run in **~1s**. Running it multiple times for various checks... huge performance impact.

Changing manually the query to:
```
SELECT * FROM `{MY_STREAM}` USE INDEX(ix_query_aggregate)
WHERE aggregate_type = '{MY_TYPE}'  AND 'aggregate_id' = '{MY_AGGR}'  AND `no` >= 1
ORDER BY `no` ASC
LIMIT 10000
```
works as expected and runs in **~0.01s**!

For reference:
https://benjaminlistwon.com/blog/working-with-json-data-in-mysql-part-1-of-3/ (**INDEXING AND EFFICIENCY**)

kochen:
fixes #147

### DIFF
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -808,13 +808,19 @@ EOT;
         }
     }
 
+    /**
+     * Convert metadata fields into indexed columns
+     * @example `_aggregate__id` => `aggregate_id`
+     */
     private function convertToColumn(array &$match): void
     {
-        $indexedColumns = $this->persistenceStrategy->indexedColumns();
-        $field = substr($match['field'], 1);
-        if (in_array($field, $indexedColumns)) {
-            $match['field'] = $field;
-            $match['fieldType'] = FieldType::MESSAGE_PROPERTY();
+        if ($this->persistenceStrategy instanceof MariaDBIndexedPersistenceStrategy) {
+            $indexedColumns = $this->persistenceStrategy->indexedColumns();
+            $field = substr($match['field'], 1);
+            if (in_array($field, $indexedColumns)) {
+                $match['field'] = $field;
+                $match['fieldType'] = FieldType::MESSAGE_PROPERTY();
+            }
         }
     }
 }
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -815,7 +815,7 @@ EOT;
     private function convertToColumn(array &$match): void
     {
         if ($this->persistenceStrategy instanceof MariaDBIndexedPersistenceStrategy) {
-            $indexedColumns = $this->persistenceStrategy->indexedMetadaColumns();
+            $indexedColumns = $this->persistenceStrategy->indexedMetadataFields();
             $field = substr($match['field'], 1);
             if (in_array($field, $indexedColumns)) {
                 $match['field'] = $field;
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -815,7 +815,7 @@ EOT;
     private function convertToColumn(array &$match): void
     {
         if ($this->persistenceStrategy instanceof MariaDBIndexedPersistenceStrategy) {
-            $indexedColumns = $this->persistenceStrategy->indexedColumns();
+            $indexedColumns = $this->persistenceStrategy->indexedMetadaColumns();
             $field = substr($match['field'], 1);
             if (in_array($field, $indexedColumns)) {
                 $match['field'] = $field;
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -31,12 +31,6 @@ use Prooph\EventStore\Util\Assertion;
 
 final class MariaDbEventStore implements PdoEventStore
 {
-    const columnsMap = [
-        '_aggregate_id' => 'aggregate_id',
-        '_aggregate_type' => 'aggregate_type',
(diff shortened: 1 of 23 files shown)


src/PersistenceStrategy.php
```diff
@@ -28,4 +28,6 @@ public function columnNames(): array;
     public function prepareData(Iterator $streamEvents): array;
 
     public function generateTableName(StreamName $streamName): string;
+
+    public function indexedColumns(): array;
```
codeliner: We should not force each PersistenceStrategy to implement this interface. To keep BC I'd say we add an addtional interface `IndexedColumnStrategy` or even MariaDB specific `MariaDbIndexedColumnStrategy`

Let only the MariaDB strategies implement that additional interface and check for it specifically.

Other than that I really like the fix :+1:

kochen: Thanks!
The only thing I see is that `MariaDbEventStore` requires a `PersistenceStrategy` which won't see the methods from  our new `MariaDbIndexedColumnStrategy` unless we extend it.

Would that be ok?


If we do so we basically say to everybody implementing a new `[MariaDBBased]PersistenceStrategy` that:
if you don't implement this new interface you will suffer from performance.

Maybe we should just put an `abstractPersistenceStrategy` that has that method implemented and returning an empty array?

src/MariaDbEventStore.php
```diff
@@ -806,4 +807,14 @@ private function createSchemaFor(string $tableName): void
             }
         }
     }
+
+    private function convertToColumn(array &$match): void
```
codeliner: Regarding MariaDbIndexedColumnStrategy
I'd just add a check here:

```php
if($this->persistenceStrategy instanceof MariaDbIndexedColumnStrategy) { ... }

Comment (codeliner) [reply src/MariaDbEventStore.php]: This way public API doesn't change but the performance bottleneck is solved internally. This should be documented, though. In the next major release we can respect the difference in the public API as well.

**Comment by kochen** [comment]
@codeliner done and updated the docs as well.

**Comment by kochen** [comment]
@prolic yes they do... (oops!)

**Comment by prolic** [comment]
`MariaDbSimpleStreamStrategy` returns an empty array of indexedColumns, so why implement this interface here at all?

**Comment by prolic** [comment]
Another thing I noticed, `no` f.e. is also an indexed column and is not mentioned in the `indexedColumns` method (and it also should not be). Hence the method name is misleading. Maybe we should rename to something more specific like `indexedMetadaFields` f.e.

**Comment by kochen** [comment]
done, thanks!

File: src/PersistenceStrategy/MariaDbAggregateStreamStrategy.php [hunk 3]
```diff
@@ -78,7 +78,7 @@ public function prepareData(Iterator $streamEvents): array
         return $data;
     }
 
-    public function indexedColumns(): array
+    public function indexedMetadaColumns(): array
```
prolic: typo: `indexedMetadaColumns`

prolic: also as metadata is json, I would prefer `fields` instead of `columns`, since json has no columns.

prolic:
Taking a look at `MariaDbSingleStreamStrategy`:

```
`aggregate_id` CHAR(36) GENERATED ALWAYS AS (JSON_UNQUOTE(JSON_EXTRACT(metadata, '$._aggregate_id'))) STORED,
```

We see metadata field is: `_aggregate_id` and the generated (indexed) column is `aggregate_id`.

`indexedMetadataFields()` method returns `aggregate_id`, so when a query comes with MetadataMacher defined like:

```
array (
    'field' => '_aggregate_id',
    'operator' => 'EQUALS',
    'value' => 'some_id',
    'fieldType' => 'METADATA,
  )
```

the metadata field is NEVER changed to the generated (indexed column). Your PR is therefore not working at all. Have you ever tried it?

I think to resolve this, we need a key => value mapping, where key = metadata field and value = generated (indexed) column name.

kochen:
@prolic that's true the `MetadataMacher` is not changed, but a single `$match` is converted upon building the WHERE clause, by simply removing the prefix `_`:
```
$field = substr($match['field'], 1);
```

Yes, I did create a mapping array in the first patch, but then I replaced it with a simpler `substr`.

And yes, it does seem to work just fine.
This is the `$where` without my converstion:
```
array(2) {
  [0]=>
  string(56) "json_value(metadata, '$._aggregate_type') = :metadata_0 "
  [1]=>
  string(54) "json_value(metadata, '$._aggregate_id') = :metadata_1 "
}
```

This is with:
```
array(2) {
  [0]=>
  string(29) "aggregate_type = :metadata_0 "
  [1]=>
  string(27) "aggregate_id = :metadata_1 "
}
```

prolic:
great!

PR #151, opened 2018-05-04
MariaDB index mapping

### LINKED ISSUE #150: Key value mapping in indexedMetadataFields required
Current implementation uses substr to map _aggregate_id to aggregate_id. Problem is, users are free to index other metadata fields as well, so they can index a field named 'house', but would need to generate a field named 'ouse'. But what if they have another field named 'mouse' that needs to be indexed as well? As you see, the substr solution was not that a good idea.

Unfortunately I released this yesterday and changing again is a BC. I would however still want to change it, because the problem is real and not solvable with current release and also the current release is a BC because of 'house' -> 'ouse' naming needs. And lf course, the release is only one day old....

Thoughts? @codeliner @kochen

kochen:
fixes #150

### DIFF
--- a/src/MariaDBIndexedPersistenceStrategy.php
+++ b/src/MariaDBIndexedPersistenceStrategy.php
@@ -17,9 +17,14 @@ interface MariaDBIndexedPersistenceStrategy
     /**
      * Return an array of indexed columns to enable the use of indexes in MariaDB
      *
-     * @return string[]
+     * @return string[][]
      *
-     * @example ['aggregate_id', 'aggregate_type', 'aggregate_version']
+     * @example
+     *      [
+     *          '_aggregate_id' => 'aggregate_id',
+     *          '_aggregate_type' => 'aggregate_type',
+     *          '_aggregate_version' => 'aggregate_version',
+     *      ]
      */
     public function indexedMetadataFields(): array;
 }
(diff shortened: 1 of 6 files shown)


src/MariaDBIndexedPersistenceStrategy.php
```diff
@@ -17,9 +17,14 @@
     /**
      * Return an array of indexed columns to enable the use of indexes in MariaDB
      *
-     * @return string[]
+     * @return string[][]
      *
```
prolic: Remove this line. It doesn't return an array of arrays

src/MariaDbEventStore.php
```diff
@@ -810,15 +810,17 @@ private function createSchemaFor(string $tableName): void
 
     /**
      * Convert metadata fields into indexed columns
+     *
+     * @param string[]
```
prolic: Param array, it's key value, not an array of strings

src/MariaDbEventStore.php
```diff
@@ -810,15 +810,17 @@ private function createSchemaFor(string $tableName): void
...
      * @example `_aggregate__id` => `aggregate_id`
      */
     private function convertToColumn(array &$match): void
     {
         if ($this->persistenceStrategy instanceof MariaDBIndexedPersistenceStrategy) {
             $indexedColumns = $this->persistenceStrategy->indexedMetadataFields();
-            $field = substr($match['field'], 1);
-            if (in_array($field, $indexedColumns)) {
-                $match['field'] = $field;
+            if (in_array($match['field'], array_keys($indexedColumns))) {
```
prolic: Add param true

prolic:
@codeliner Are you okay with this as well? I like to merge and release this asap, and don't want to mess it up again.

PR #152, opened 2018-05-05
Fix index metadata in MariaDb single stream strategy

stepiiik:
There was a typo.


prolic:
thanks!

kochen:
Thanks @stepiiik

Issue #157, opened 2018-05-24
Postgres event store with schema support


ghettovoice:
Hello,
thank you for the awesome library! 
I use pdo store with postgres and I want to create store tables in the custom DB schema.
Is it planned to add support of multi schemas event stores in PostrgesEventStore and Persistence strategies? Currently this is impossible due to quoting the whole $tableName variable.
Thanks!

ghettovoice:
I need this feature in my current working project where we have separate schema for each module.
So I've done a pull request with changes to Pg related classes to add such feature.
Seems like all work as before, tests are passed. Le me know if you have questions about pull request.
Thank you for the great job!

PR #159, opened 2018-05-25
Pg schema support

ghettovoice:
This pull request adds support of multi schema events stores and projection for PostresSQL implementation #157 .
Summary (only Pg related classes):
- Event streams  table and projections table name can be defined with custom DB schema.
- String before first dot in stream name treated as DB schema, so stream auto generated table will be created in custom schema.
I added additional tests to clarify the new behaviour

### DIFF
--- a/src/PersistenceStrategy/PostgresAggregateStreamStrategy.php
+++ b/src/PersistenceStrategy/PostgresAggregateStreamStrategy.php
@@ -16,17 +16,22 @@ use Iterator;
 use Prooph\EventStore\Pdo\Exception;
 use Prooph\EventStore\Pdo\PersistenceStrategy;
 use Prooph\EventStore\StreamName;
+use Prooph\EventStore\Pdo\Util\PostgresHelper;
 
 final class PostgresAggregateStreamStrategy implements PersistenceStrategy
 {
+    use PostgresHelper;
+
     /**
      * @param string $tableName
      * @return string[]
      */
     public function createSchema(string $tableName): array
     {
+        $tableName = $this->quoteIdent($tableName);
+
         $statement = <<<EOT
-CREATE TABLE "$tableName" (
+CREATE TABLE $tableName (
     no BIGSERIAL,
     event_id UUID NOT NULL,
     event_name VARCHAR(100) NOT NULL,
@@ -40,7 +45,7 @@ EOT;
 
         return [
             $statement,
-            "CREATE UNIQUE INDEX on \"$tableName\" ((metadata->>'_aggregate_version'));",
+            "CREATE UNIQUE INDEX on $tableName ((metadata->>'_aggregate_version'));",
         ];
     }
 
@@ -78,6 +83,9 @@ EOT;
 
     public function generateTableName(StreamName $streamName): string
     {
-        return '_' . sha1($streamName->toString());
+        return implode('.', array_filter([
+            $this->extractSchema($streamName->toString()),
+            '_' . sha1($streamName->toString()),
+        ]));
     }
 }
--- a/src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
+++ b/src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
@@ -15,17 +15,21 @@ namespace Prooph\EventStore\Pdo\PersistenceStrategy;
 use Iterator;
 use Prooph\EventStore\Pdo\PersistenceStrategy;
 use Prooph\EventStore\StreamName;
+use Prooph\EventStore\Pdo\Util\PostgresHelper;
 
 final class PostgresSimpleStreamStrategy implements PersistenceStrategy
 {
+    use PostgresHelper;
     /**
      * @param string $tableName
      * @return string[]
      */
     public function createSchema(string $tableName): array
     {
+        $tableName = $this->quoteIdent($tableName);
+
         $statement = <<<EOT
-CREATE TABLE "$tableName" (
+CREATE TABLE $tableName (
     no BIGSERIAL,
     event_id UUID NOT NULL,
     event_name VARCHAR(100) NOT NULL,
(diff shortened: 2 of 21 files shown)


src/Util/PostgresHelper.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Prooph\EventStore\Pdo\Util;
+
+trait PostgresHelper {
+    /**
+     * @param string $tableName
+     * @return string
+     */
+    private function quoteIdent(string $tableName): string
+    {
+        return array_reduce(explode('.', $tableName), function ($result, $part) {
```
prolic: that's really hard to read, can we simplify that?

src/Util/PostgresHelper.php
```diff
@@ -0,0 +1,33 @@
...
+        }, '');
+    }
+
+    /**
+     * @param string $identifier
+     * @return string|null
+     */
+    private function extractSchema(string $identifier): ?string
+    {
+        $parts = array_map(function ($part) {
```
prolic: same here

tests/Assets/scripts/postgres/01_custom_event_streams_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SCHEMA IF NOT EXISTS custom;
```
prolic: I don't know if "custom" is a good default, why not "prooph" :smile:

ghettovoice:
Thank you for review! I'll fix it.
As I know in MySQL/MariaDB `schema` represents another database and probably with another connection configuration. Currently I don't know how to deal with this

prolic:
You're right, thanks. So no worries.

Can you also please make sure that existing app configurations are not
broken and the use of additional schema is optional?

src/Util/PostgresHelper.php
ghettovoice: ready

src/Util/PostgresHelper.php
ghettovoice: ready

tests/Assets/scripts/postgres/01_custom_event_streams_table.sql
ghettovoice: ready

ghettovoice:
> Can you also please make sure that existing app configurations are not
> broken and the use of additional schema is optional?

Tables for event streams and projections will be created in custom schema only if `$eventStreamsTable`, `$projectionsTable` and stream name will contain dot and only for Postgres driver. I highly doubt anyone is using a point in these variables intentionally right now, so there should't be problems.

prolic:
May I asked why it's a requirement to have schemas in your project?

ghettovoice:
Sure, it just an our internal convention: not to create all at public schema (pg extensions usually creates here functions, tables and etc.), each module adds tables, functions and etc in it's  own schema, so it easy to search in relatively large.

ghettovoice:
Ok , thank you

prolic:
I'll have to make some more testing, I don't like this quoting method, I hope to come up with a faster solution. Currently really busy.

prolic:
@ghettovoice I think I found a much faster quoting-function, see https://3v4l.org/7bCYi

ghettovoice:
@prolic I see :+1: 
My implementation uses two functions because of needless schema extraction in `generateTableName` 
method https://github.com/prooph/pdo-event-store/pull/159/commits/e1b9a7b440656f4d846dd7eded5dcff06c33f455#diff-6895a5b565f06c33a1404ad8257df13e.
May be here should be another solution for persistence strategy?

ghettovoice:
Ok, I'll try to make this before weekends

src/PersistenceStrategy/PostgresAggregateStreamStrategy.php
```diff
@@ -83,11 +83,14 @@ public function prepareData(Iterator $streamEvents): array
...
+        $streamName = $streamName->toString();
+        $table = '_' . sha1($streamName);
+        [$schema] = $this->splitIdent($streamName);
 
-        return implode('.', array_filter([
-            $schema,
-            '_' . sha1($streamName->toString()),
-        ]));
+        if ($schema) {
+            $schema . '.' . $table;
```
prolic: That doesn't make any sense

src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
```diff
@@ -75,11 +75,14 @@ public function prepareData(Iterator $streamEvents): array
...
+        $streamName = $streamName->toString();
+        $table = '_' . sha1($streamName);
+        [$schema] = $this->splitIdent($streamName);
 
-        return implode('.', array_filter([
-            $schema,
-            '_' . sha1($streamName->toString()),
-        ]));
+        if ($schema) {
+            $schema . '.' . $table;
```
prolic: same here

src/PersistenceStrategy/PostgresSingleStreamStrategy.php
```diff
@@ -90,11 +90,14 @@ public function prepareData(Iterator $streamEvents): array
...
+        $streamName = $streamName->toString();
+        $table = '_' . sha1($streamName);
+        [$schema] = $this->splitIdent($streamName);
 
-        return implode('.', array_filter([
-            $schema,
-            '_' . sha1($streamName->toString()),
-        ]));
+        if ($schema) {
+            $schema . '.' . $table;
```
prolic: and again, maybe you should add a unit test for this method

ghettovoice:
Now it's done. Yesterday I made it late at night

src/Util/PostgresHelper.php
```diff
@@ -0,0 +1,54 @@
...
+
+        return '"' . $schema . '"."' . $table . '"';
+    }
+
+    /**
+     * Splits fully qualified table name by first dot.
+     * @param string $ident
+     * @return string[]
+     */
+    private function splitIdent(string $ident): array
```
prolic: On reviewing I found that the returned table name is not used at all. So why not change the signature then? Something like `private function schema(string $ident): ?string`. It either returns the schema or null

ghettovoice: There was `private function extractSchema(string $ident): ?string` in the very first version.
I'll fix this

ghettovoice:
Thank you for your time!

PR #162, opened 2018-06-07
Check if projection exists before creating it

### LINKED ISSUE #160: Ignored error in PdoEventStoreReadModelProjector
In `PdoEventStoreReadModelProjector` there is a case where an SQL query that causes an error is intentionally ignored ([here](https://github.com/prooph/pdo-event-store/blob/c3e925cf1c8b169b753e363a602a728e1e1c1aee/src/Projection/PdoEventStoreReadModelProjector.php#L682-L693)).

While this error indeed looks like something that can be ignored, it does actually cause some problems.

First in our import we're running `PdoEventStoreReadModelProjector::run()` a lot of times which causes our log for Postgres docker container to be full of this ignored error, making it difficult to debug other issues.

Second problem is that in some obscure cases with transaction handling (we're trying to wrap each of our tests in a transaction and revert it afterwards) it causes the next SQL query to fail with this error:

```
Prooph\EventStore\Exception\RuntimeException: Error 25P02. Maybe the projection table is not setup?
      Error-Info: ERROR:  current transaction is aborted, commands ignored until end of transaction block in vendor/prooph/pdo-event-store/src/Projection/PdoEventStoreReadModelProjector.php:722
```

When I added `ON CONFLICT (name) DO NOTHING` to the insert in `createProjection` both of the problems above disappeared.

Of course `ON CONFLICT` is a Postgres 9.5+ specific syntax so I can't send a PR with that.

Since "upsert" syntax is different for each database and in some cases doesn't exist at all, the only cross-platform fix is to check with a select before running the insert. In my opinion this is alright to do since `createProjection` is only called once when you run the projections.

If you're alright with the suggested fix, I can send a PR.

enumag:
Closes #160.

I can't really think of a good test for this since the end result is exactly the same except database logs. However I've tried some on purpose broken implementations and the tests failed every time so I think we're covered.

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -41,11 +41,6 @@ final class PdoEventStoreProjector implements Projector
         extractSchema as pgExtractSchema;
     }
 
-    private const UNIQUE_VIOLATION_ERROR_CODES = [
-        'pgsql' => '23505',
-        'mysql' => '23000',
-    ];
-
     /**
      * @var EventStore
      */
@@ -492,7 +487,10 @@ EOT;
                 break;
         }
 
-        $this->createProjection();
+        if (! $this->projectionExists()) {
+            $this->createProjection();
+        }
+
         $this->acquireLock();
 
         $this->prepareStreamPositions();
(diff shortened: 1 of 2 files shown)


prolic:
Same applies here: PdoEventStoreProjector

PR #163, opened 2018-06-07
Add message to ConcurrencyException

### LINKED ISSUE #161: Confusing ConcurrencyException
Recently we encountered some cases of `Prooph\EventStore\Exception\ConcurrencyException` (thrown [here](https://github.com/prooph/pdo-event-store/blob/c3e925cf1c8b169b753e363a602a728e1e1c1aee/src/PostgresEventStore.php#L221)).

From the name we thought the EventStore failed because there was some other process already working with it, but that wasn't the case. After I added some debug code to dump `$statement->errorInfo()` the cause was more clear:

```
array(3) {
  [0] =>
  string(5) "23505"
  [1] =>
  int(7)
  [2] =>
  string(355) "ERROR:  duplicate key value violates unique constraint "_c1455064a23e9da532a1f0248b49e1d402af81f4_expr_expr1_expr2_idx"
DETAIL:  Key ((metadata ->> '_aggregate_type'::text), (metadata ->> '_aggregate_id'::text), (metadata ->> '_aggregate_version'::text))=(App\RealtyRegistration\DomainModel\Realty, e2eb56d1-4705-4210-acb7-39112a4f69c5, 1) already exists."
}
````

Based on that I found out that we had an error in our testing data which caused one UUID to be accidentally used for two aggregates - trying to insert the second one caused the exception.

While this was entirely our fault, I have a problem with the exception.

- First the name suggests a different cause than the actual problem.
- Second there is no message so you don't even know the UUID of the failing aggregate without debugging.
- Third the exception is caught in `ActionEventEmitterEventStore` and thrown again later, but it's a new instance so the stack trace of the original exception is lost.

I don't have any idea for improving the third point but we can surely do something about the first two. However since the exception is triggered by two different database error codes (`23000` and `23505`) I'm not entirely sure if the issue I encountered is the only one that can trigger it. That's why I'd like to discuss this first before sending a PR. Can you give me some insight from your perspective for what this exception is about and why is it named the way it is?

### PR DESCRIPTION
**Pull request by enumag**
Closes #161.

I don't think we can easily detect if the problem was an aggregate ID or an event ID (not without an additional select to check for duplicates). However the information seems to be in PDO's ErrorInfo so I'd like to simply pass it along in the exception message.

This will of course require another PR for `ActionEventEmitterEventStore` to pass the message along [here](https://github.com/prooph/event-store/blob/master/src/ActionEventEmitterEventStore.php#L88-L89) instead of the boolean.

I'll send that PR later. Unless of course you're against this solution.

### DIFF
--- a/src/Exception/ConcurrencyExceptionFactory.php
+++ b/src/Exception/ConcurrencyExceptionFactory.php
@@ -1,0 +1,19 @@
+<?php declare(strict_types = 1);
+
+namespace Prooph\EventStore\Pdo\Exception;
+
+use Prooph\EventStore\Exception\ConcurrencyException;
+
+class ConcurrencyExceptionFactory
+{
+    public static function fromStatementErrorInfo(array $errorInfo): ConcurrencyException
+    {
+        return new ConcurrencyException(
+            sprintf(
+                "Some of the aggregate IDs or event IDs have already been used in the same stream. The PDO error should contain more information:\nError %s.\nError-Info: %s",
+                $errorInfo[0],
+                $errorInfo[2]
+            )
+        );
+    }
+}
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -17,12 +17,12 @@ use Iterator;
 use PDO;
 use PDOException;
 use Prooph\Common\Messaging\MessageFactory;
-use Prooph\EventStore\Exception\ConcurrencyException;
 use Prooph\EventStore\Exception\StreamExistsAlready;
 use Prooph\EventStore\Exception\StreamNotFound;
 use Prooph\EventStore\Metadata\FieldType;
 use Prooph\EventStore\Metadata\MetadataMatcher;
 use Prooph\EventStore\Metadata\Operator;
+use Prooph\EventStore\Pdo\Exception\ConcurrencyExceptionFactory;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/Exception/ConcurrencyExceptionFactory.php [hunk 1]
```diff
@@ -0,0 +1,19 @@
+<?php declare(strict_types = 1);
+
+namespace Prooph\EventStore\Pdo\Exception;
+
+use Prooph\EventStore\Exception\ConcurrencyException;
+
+class ConcurrencyExceptionFactory
```
prolic: This can be a named constructor on the concurrency exception itself, no need for an additional class

prolic:
About action event emitter, why not set the exception itself instead if bool true?

Same could be for stream not found

src/Exception/ConcurrencyExceptionFactory.php
enumag: It can't. ConcurrencyException is in another package which should not be aware of something like PDO.

The only way I could use a named constructor here would be to extend that exception. I slightly prefer this solution though.

prolic: Oh you're right. I'm okay with having a factory then.

enumag:
@prolic Oh I somehow thought the event parameters are limited to just scalars and maybe arrays of scalars. If that's not the case then just setting the exception to the parameter is of course far better since it will preserve the original call stack.

Issue #164, opened 2018-06-29
MariaDB 10.2.16 causes issues


codeliner:
During release of pdo-event-store 1.9 we encountered errors testing mariadb.
BETA was tested against mariadb 10.2.15 and everything worked fine. But now with mariadb 10.2.16 being available our travis build fails: https://travis-ci.org/prooph/pdo-event-store/jobs/398146743

We were able to reproduce the bug locally using official mariadb docker images:
mariadb 10.2.15 :heavy_check_mark: 
mariadb 10.2.16 :no_entry_sign: 

**We try to identify the exact problem. For now, don't use mariadb 10.2.16 with pdo-event-store!!!**

gquemener:
Here are the steps to reproduce the phpunit errors locally:

``` bash
$ git clone git@github.com:prooph/pdo-event-store.git
$ cd pdo-event-store
$ docker run --rm -ti -v "$PWD":/app --user "$(id -u):$(id -g)" -w /app composer install --ignore-platform-reqs
$ docker run --rm -ti -p "3306:3306" -e MYSQL_ALLOW_EMPTY_PASSWORD=yes --name pdo-event-store-mariadb -d mariadb:10.2.16
$ docker exec -it pdo-event-store-mariadb mysql -e 'create database event_store_tests;'
$ cp phpunit.xml.mariadb phpunit.xml
$ sed -i 's/127.0.0.1/pdo-event-store-mariadb/g' phpunit.xml
$ docker run --rm -ti -v "$PWD":/app -w /app --user "$(id -u):$(id -g)" --link "pdo-event-store-mariadb" prooph/php:7.1-cli vendor/bin/phpunit
$ docker kill pdo-event-store-maria-db
```

And here's the diff between https://github.com/MariaDB/server/compare/mariadb-10.2.15...mariadb-10.2.16 (if of any help).

basz:
remark. I found I needed 10.2.14 not 10.2.15 to get the tests working. So, 10.2.14 one would need.

drgomesp:
I can confirm what @basz says, as I also ran into the same errors with **10.2.15**.

prolic:
Failed asserting that 5 matches expected 49


That's because of a bugfix in eventstore I guess. See latest release

basz:
MariaDB version | Result
------------ | -------------
mariadb:10.2.14 | OK (394 tests, 2183 assertions)
mariadb:10.2.15 | OK (394 tests, 2183 assertions)
[...]

PR #165, opened 2018-07-12
Start of batch size bug report.

bweston92:


bweston92:
@prolic is this a bug or am I just being stupid?

prolic:
at least tests are failing now. I'll try to investigate asap.

basz:
I've identified 3 issues

1. PR related.
```
1) ProophTest\EventStore\Pdo\MySqlCustomStrategiesEventStoreTest::it_loads_correctly_using_single_stream
Failed asserting that 3 is equal to 2 or is less than 2.
```

2. Codestyle issues. Did we switch from `somephpfunction()` to `\somephpfunction()`?
easily fixed by running `composer cs-fix`.

3. Issues with the maria adapter. 8x `BadMethodCallException: Accessing the key of an EmptyIterator`. Probably due to upgraded packages via `composer update`.

prolic:
MariaDB is not a problem. This is due to a change in newest MariaDB server (they don't know what a BC is).

See https://github.com/prooph/pdo-event-store/issues/164

basz:
Should pin MariaDB then?

basz:
@prolic Are you sure? I see the same test failures with `image: mariadb:10.2.15`. 

```
2) ProophTest\EventStore\Pdo\MariaDbCustomStrategiesEventStoreTest::it_loads_events_by_matching_metadata with data set #3 (array(true, 123, 'bar'))
BadMethodCallException: Accessing the key of an EmptyIterator

/app/vendor/prooph/event-store/tests/AbstractEventStoreTest.php:365
```

bweston92:
This is just a logic issue in the code setting the wrong limit.

prolic:
I would like to get this into next release. Can we make this work? @bweston92 @basz

basz:
@bweston92 do a rebase pls, before continuing. Think only your primary concern is left to solve.

basz:
> This is just a logic issue in the code setting the wrong limit.

From what I can tell I'm not sure setting the batch size should in 2 loaded events on the iterator. Setting the batch size just means the adaptor will use LIMIT in its queries, but this is an implementation detail that should not affect the count on the iterator.

@prolic can you confirm the store should work as I described?

If so, this issue seems resolved although a count method is nice to have. The test is then simply not needed.

basz:
I think I see the issue here: https://github.com/prooph/pdo-event-store/blob/2f5e4f85a13f36b2dd835e9ed1beec6c4cf02419/src/PdoStreamIterator.php#L238 not sure how to fix it yet...

bweston92:
Yes the wrong limit is calculated.

basz:
i see now, lets see if https://github.com/prooph/pdo-event-store/pull/171 solves it

src/PdoStreamIterator.php
```diff
@@ -228,6 +228,11 @@ public function rewind(): void
         }
     }
 
+    public function count(): int
```
prolic: There is a problem with this approach. It will return count only for current batch size, not count for the iterator. This can lead to users using this method and receiving unexpected results.

Can we make it private and call it via reflection api in the tests?

/cc @bweston92 @basz

prolic:
I think you can remove count method and use reflection only in the tests.

Issue #168, opened 2018-08-26
deleting running projections will not restart from event 0


basz:
When I have a projection running that is instructed to delete with deletion projections are not removed until *all* events have been processed. 

I would expect it would finish projecting events within the batch size configuration option to then delete and restart.

discovered this via snapshots which are also read model projections.

Looking at the code this seems to be true for all status changes (stop, delete and reset) as the method `handleStreamWithSingleHandler` does not check for status at all.

I've experimented a bit to fix this inside the `handleStreamWithSingleHandler` method, but it's not as simple as calling this->reset(). One possible solution seems to be to simply stop when status is not running.

prolic:
@basz can this be closed now?

basz:
Closing per #170

PR #170, opened 2018-08-26
Fix/not resetting until all done

basz:
see https://github.com/prooph/pdo-event-store/issues/168

test in parent repo will fail. I believe as expected...

I would need help writing a test for this...

### DIFF
--- a/.travis.install-mariadb-10.2.14.sh
+++ b/.travis.install-mariadb-10.2.14.sh
@@ -1,0 +1,5 @@
+sudo apt-get install software-properties-common
+sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xcbcb082a1bb943db
+sudo add-apt-repository 'deb https://downloads.mariadb.com/MariaDB/mariadb-10.2.14/repo/ubuntu trusty main'
+sudo apt-get update -q
+sudo apt-get install -q -y --force-yes -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" mariadb-server
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
         - DRIVER="pdo_mysql"
         - DB=mariadb_10.2
       addons:
-          mariadb: '10.2'
+          mariadb: '10.2.15'
     - php: 7.1
       dist: trusty
       sudo: true
@@ -22,7 +22,7 @@ matrix:
         - DRIVER="pdo_mysql"
         - DB=mariadb_10.2
       addons:
-          mariadb: '10.2'
+          mariadb: '10.2.15'
     - php: 7.1
       dist: trusty
       sudo: true
@@ -32,7 +32,7 @@ matrix:
         - DB=mariadb_10.2
         - DB_ATTR_ERRMODE=2 # \PDO::ERRMODE_EXCEPTION
       addons:
-          mariadb: '10.2'
+          mariadb: '10.2.15'
     - php: 7.1
       sudo: true
       env:
@@ -134,7 +134,7 @@ matrix:
         - DRIVER="pdo_mysql"
         - DB=mariadb_10.2
       addons:
-          mariadb: '10.2'
+          mariadb: '10.2.15'
     - php: 7.2
       dist: trusty
       sudo: true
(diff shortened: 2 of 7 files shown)


prolic:
Can you rebase please? I can't read that diff

basz:
rebased

prolic:
For test case: you can append 5000 events, have batch size 500. After 1944 events (counting in function) you call stop. See if behavior is expected.

basz:
Would you do that with an internal projection or with an isolated one?

prolic:
I don't get the question

prolic:
Actually doesn't matter I think

basz:
ok.

do you agree with me on this? (assuming you agree with this change at all)

> That last line (L466) should actually have an expectation of 5, not 50.

prolic:
Why 5? Why 50? UsernameChanged events happens 49 times, see https://github.com/prooph/event-store/blob/master/tests/Projection/AbstractEventStoreProjectorTest.php#L1008

basz:
5 because of https://github.com/prooph/event-store/blob/master/tests/Projection/AbstractEventStoreProjectorTest.php#L482

With my proposed change stopping/resetting/deleting a running or idle projection would stop it and the projected events should be the persist_block_size option
 
i meant the expectation of 49 should be 5

prolic:
Oh, probably yes, seems like you are right.

prolic:
@basz Thinking about this: I think this is already the test-case, you don't need to write one. Just fix the test-case and make it pass.

basz:
Thats what I was thinking. Its located in prooph/event-store.

basz:
@prolic I've update the tests in prooph/event-store. Could you please restart travis to see if it works?  it does here locally.

basz:
(after merge, come to think of it)

basz:
should I add a minimum dependency of `event-store:7.3.8` to composer? Otherwise the travis runs with minimum requirements will fail.

prolic:
event-store 7.3.8 released, please pin this in composer json

basz:
ugg, should i also pin the mariadb image, so we don't have those pesky errors there?

prolic:
yes please :+1:

prolic:
Mariadb tests are all failing.

basz:
I know; but do you have an idea why? How come JSON queries are syntax erroring. (as said work locally with docker)

prolic:
From travis log file:

```
E: Failed to fetch http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.2.14/ubuntu/dists/trusty/main/binary-amd64/Packages  404  Not Found
E: Failed to fetch http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.2.14/ubuntu/dists/trusty/main/binary-i386/Packages  404  Not Found
E: Some index files failed to download. They have been ignored, or old ones used instead.
$ PACKAGES='mariadb-server mariadb-server-10.2.14'
$ if [[ $(lsb_release -cs) = 'precise' ]]; then PACKAGES="${PACKAGES} libmariadbclient-dev"; fi
0.49s$ sudo apt-get install -y -o Dpkg::Options::='--force-confnew' $PACKAGES
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package mariadb-server-10.2.14
E: Couldn't find any package by glob 'mariadb-server-10.2.14'
E: Couldn't find any package by regex 'mariadb-server-10.2.14'
```

prolic:
see that file for travis to install mysql 5.7? we need something like this for mariadb, help can be found here: https://mariadb.com/kb/en/library/installing-mariadb-deb-files/

prolic:
Errors were encountered while processing:
 /var/cache/apt/archives/mariadb-server-5.5_5.5.59-1ubuntu0.14.04.1_amd64.deb
W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Sub-process /usr/bin/dpkg returned an error code (1)
The command "if [[ $DB == 'mariadb_10.2' ]]; then bash .travis.install-mariadb-10.2.14.sh; fi" failed and exited with 100 during .

basz:
yup, got it! 'ubuntu' not 'debian'

basz:
oh btw. I do wonder why we couldn't just fix our code to work with mariadb 10.2.15+. Seems kind of weird to use an old version that has security issues. But that's for another PR I guess. No one complained about this yet so far?

prolic:
Nope, but as you said, can be done in another PR. Hopefully in a way, that both are supported.

PR #171, opened 2018-09-03
Bugreport/batch size

basz:
fix #165

### DIFF
--- a/src/PdoStreamIterator.php
+++ b/src/PdoStreamIterator.php
@@ -228,11 +228,6 @@ final class PdoStreamIterator implements Iterator
         }
     }
 
-    public function count(): int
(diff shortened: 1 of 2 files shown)


src/PdoStreamIterator.php
```diff
@@ -228,14 +228,19 @@ public function rewind(): void
         }
     }
 
+    public function count(): int
```
prolic: There is a problem with this approach. It will return count only for current batch size, not count for the iterator. This can lead to users using this method and receiving unexpected results.

Can we make it private and call it via reflection api in the tests?

/cc @bweston92 @basz

basz:
like so?

PR #173, opened 2018-09-21
Removed usage of mariadb json_value with boolean

gquemener:
Prior to MariaDB 10.2.16, json_value would return the `true` or `false`
string when fetching a boolean json data. From 10.2.16, it returns the
`1` or `0` integer.

See https://jira.mariadb.org/browse/MDEV-15905
See https://github.com/prooph/pdo-event-store/issues/164

The solution provided here is to use `json_extract` which behavior hasn't changed (it always return `0` or `1` when fetching boolean json data).

--

Side Note: The testsuite passes with MariaDB 10.3.9 (locally), I could work on adding it to the travis build if that's something you'd be interested in @prolic @codeliner ?

### DIFF
--- a/.travis.install-mariadb-10.2.14.sh
+++ b/.travis.install-mariadb-10.2.14.sh
@@ -1,5 +1,0 @@
-sudo apt-get install software-properties-common
-sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xcbcb082a1bb943db
-sudo add-apt-repository 'deb https://downloads.mariadb.com/MariaDB/mariadb-10.2.14/repo/ubuntu trusty main'
-sudo apt-get update -q
-sudo apt-get install -q -y --force-yes -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" mariadb-server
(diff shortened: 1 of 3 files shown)


prolic:
Yes please, that's awesome

prolic:
Soooo nice! Thanks!

Next thing, support MySQL v8 and PostgreSQL v10 (separate PR) - wanna do this as well?

gquemener:
@basz Do you remember why you've provided a custom script for installing mariadb 10.2.14 instead of relying on the `addons` travis config) [here](https://github.com/prooph/pdo-event-store/commit/56a4fc1db16b2bbe288536b052e2bedd679397ff)

gquemener:
> Soooo nice! Thanks!

You're welcome :)

> Next thing, support MySQL v8 and PostgreSQL v10 (separate PR) - wanna do this as well?

Why not :thinking:

prolic:
@gquemener the script was added because travis was failing with newer mariadb versions. With this patched now, I think we can remove it again.

basz:
> @gquemener the script was added because travis was failing with newer mariadb versions

Yes

gquemener:
All set, ready for merge :+1:

prolic:
Great work! Thank you!

Issue #154, opened 2018-05-15
Floats in events are serialized as integers


enumag:
When I have a float value in an event without decimals (for example `1.0`) it is saved to the EventStore as simply `1`. Because of this when the event is later retrieved from the EventStore and deserialized, the information that the number was a float is lost - it becomes an integer.

If this is intended for Prooph then just close this issue. If you want to fix it then it's easy to do so using `JSON_PRESERVE_ZERO_FRACTION`. I can send a pull request for this.

----

Another thing to consider is checking if the json serialization didn't produce any errors using the `json_last_error()` function.

Personally I'm using [`Nette\Utils\Json`](https://api.nette.org/2.4/source-Utils.Json.php.html#25-47) in my project which also adds `JSON_UNESCAPED_UNICODE` and `JSON_UNESCAPED_SLASHES` flags handles some other issues. There is something similar although less advanced in [Enqueue](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Util/JSON.php#L45-L57) library which uses just `JSON_UNESCAPED_UNICODE` and `json_last_error()`.

I assume you don't want to add a dependency to another library so I won't send a pull request for this.

prolic:
We can check for json_last_error() and use additional flags, wanna submit a PR?

enumag:
Only if it would be integration of an existing library. I don't have time to copy the logic to prooph and write all necessary tests.

basz:
## summerizing

### JSON_PRESERVE_ZERO_FRACTION
definitely an issue. A value as float(-1.0) will become int(1) without 

### JSON_UNESCAPED_UNICODE
If I understand this correctly. "က" is encoded as "\u1000" without the flag en "က" with. json_decode works regardless. So adding this flag we would go from  \u1000 notation which should work in any table/column to "က" for which the correct collation should be set. Seems to me adding it (now) will be more problematic then leaving it off. Not sure. see: https://3v4l.org/FKj8I

### JSON_UNESCAPED_SLASHES
Same as JSON_UNESCAPED_UNICODE. It will change the way payloads are stored. See. https://3v4l.org/5cROr

I worry (just a bit) that changing json encoding midway might lead to problems with ~half of your events later on. Especially when someone is using the payloads directly.

### json_last_error()
[...]

PR #175, opened 2018-09-23
Better json serialization

basz:
investigating #154


prolic:
@basz @fritz-gerneth @darrylhein @enumag @gquemener anyone willing to supply a patch to make those tests pass without BC?

enumag:
We can simply add JSON_PRESERVE_ZERO_FRACTION and maybe some of the other flags and close that issue. That way we would ignore the possibilities of `json_last_error()` but that's not as important. I only posted it as a side note. I'll make sure to fix that in event-store-client when I get to use it.

basz:
discussion accidentally continued here: https://github.com/basz/pdo-event-store/pull/2

PR #177, opened 2018-11-01
Implement a second optional MetadataMatcher parameter for fromStream methods

fjogeleit:
Implementation a second optional parameter for fromStream Methods. So its possible to filter Streams for Projections, Queries and so on. As example is it possible to filter for an AggregationID to create Projections for a single Aggregate with the SingleStream- or SimpleStreamStrategy
- related to: https://github.com/prooph/event-store/pull/350


sandrokeil:
@prolic This should also be part of the appropriate interface like [`Query`](https://github.com/prooph/event-store/blob/master/src/Projection/Query.php#L29) or?

prolic:
@sandrokeil We can't change the interface, because that would be a BC.

PR #185, opened 2018-11-13
Issue/payload encoding

prolic:
superseeds https://github.com/prooph/pdo-event-store/pull/175

### DIFF
--- a/src/Util/Json.php
+++ b/src/Util/Json.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * This file is part of the prooph/pdo-event-store.
+ * (c) 2016-2018 prooph software GmbH <contact@prooph.de>
+ * (c) 2016-2018 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo\Util;
+
+use Prooph\EventStore\Pdo\Exception\JsonException;
+
+class Json
+{
+    /**
+     * @param mixed $value
+     *
+     * @return string
+     *
+     * @throws JsonException
+     */
+    public static function encode($value): string
+    {
+        $flags = \JSON_UNESCAPED_UNICODE | \JSON_UNESCAPED_SLASHES | \JSON_PRESERVE_ZERO_FRACTION;
+
+        $string = \json_encode($value, $flags);
+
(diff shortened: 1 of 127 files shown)


prolic:
Some docs are still missing to point out the MySQL bug. @basz wanna take over?

basz:
Yes

src/Util/Json.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @throws JsonException
+     */
+    public static function encode($value): string
+    {
+        $flags = \JSON_UNESCAPED_UNICODE | \JSON_UNESCAPED_SLASHES | \JSON_PRESERVE_ZERO_FRACTION;
+
+        $string = \json_encode($value, $flags);
+
+        if ($error = \json_last_error()) {
```
gquemener: This is working only because value of `JSON_ERROR_NONE` is set to 0, which evalutates to `false`.

I'd vote for explicitness with `JSON_ERROR_NONE !== $error = \json_last_error()`.

src/Util/Json.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * @param mixed $value
+     *
+     * @return string
+     *
+     * @throws JsonException
+     */
+    public static function encode($value): string
+    {
+        $flags = \JSON_UNESCAPED_UNICODE | \JSON_UNESCAPED_SLASHES | \JSON_PRESERVE_ZERO_FRACTION;
```
gquemener: What about using [`JSON_FORCE_OBJECT`](http://php.net/manual/en/json.constants.php#constant.json-force-object) as well?

src/Util/Json.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $string = \json_encode($value, $flags);
```
gquemener: I'd favor a more explicit name (`$string` may contain `false`), like `$json`, `$value`, `$data` or `$result`.

gquemener:
Thanks for taking care of this issue @basz :+1:

gquemener:
I've never used an `array` as a payload, is there really a use case?

I was more focused on encoding of an empty payload which would result in `[]` instead of `{}`.

prolic:
{'userlist':[1,4,5]} would be {'userlist':{'0':1, '1':4,'2':5}}

gquemener:
Ah yeah right, that would not be desirable ^^

Sorry for the noise, then :D

*Gildas Quéméner*

prolic:
Nice, goog work @basz

basz:
Gery welcome

PR #188, opened 2019-01-11
Fix streamCreated condition

rodion-k:


prolic:
Can you provide a test case that is failing without that patch?

rodion-k:
I have added tests. it_should_not_create_projection_stream_on_emit_if_it_already_exists fail without patch

prolic:
thanks

PR #190, opened 2019-02-07
Feature/189 write locks

fritz-gerneth:


docs/variants.md
```diff
@@ -121,6 +121,24 @@ transactions.
 Note: This could lead to problems using the event store, if you did not manage to handle the transaction handling accordingly.
 This is your problem and we will not provide any support for problems you encounter while doing so.
 
+### Using write locks
+
+In high-concurrent write scenarios it is possible that events are skipped when reading from the stream simultaneously. 
+[Issue #198](https://github.com/prooph/pdo-event-store/issues/189) explains the background in more detail on why this
```
alle: A typo, here:
```suggestion
[Issue #189](https://github.com/prooph/pdo-event-store/issues/189) explains the background in more detail on why this 
```

fritz-gerneth: Thanks :)

fritz-gerneth:
I have no idea why those last tests are actually failing... ?

docs/variants.md
```diff
@@ -121,6 +121,24 @@ transactions.
...
 This is your problem and we will not provide any support for problems you encounter while doing so.
 
+### Using write locks
+
+In high-concurrent write scenarios it is possible that events are skipped when reading from the stream simultaneously. 
+[Issue #189](https://github.com/prooph/pdo-event-store/issues/189) explains the background in more detail on why this 
+is happening. 
+
+In order to prevent this it is possible to apply a locking strategy when writing to an event stream. 
+By default the `NoLockStragey` does not conduct any form of locking. the `MysqlMetadataLockStrategy` uses MySQL
```
bwowk:
```suggestion
By default the `NoLockStrategy` does not conduct any form of locking. the `MysqlMetadataLockStrategy` uses MySQL
```
typo here

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
```diff
@@ -0,0 +1,75 @@
...
+    private $connection;
+
+    /**
+     * @var int
+     */
+    private $timeout;
+
+    public function __construct(\PDO $connection, int $timeout = 0xffffff)
+    {
+        if ($timeout < 0) {
```
bwowk:
```suggestion
        if ($timeout <= 0) {
```

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
```diff
@@ -0,0 +1,75 @@
...
+            throw new \InvalidArgumentException('$timeout must be larger 0.');
```
bwowk:
```suggestion
            throw new \InvalidArgumentException('$timeout must be greater than 0.');
```

prolic:
> I have no idea why those last tests are actually failing... ?

travis hickup, i restarted those builds

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
```diff
@@ -0,0 +1,75 @@
...
+     * @var \PDO
+     */
+    private $connection;
+
+    /**
+     * @var int
+     */
+    private $timeout;
+
+    public function __construct(\PDO $connection, int $timeout = 0xffffff)
```
prolic: Aint that timeout so high, that it doesn't even make sense to timeout at all? It times out after 6 month.

src/WriteLockStrategy/MysqlMetadataLockStrategy.php
```diff
@@ -0,0 +1,67 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo\WriteLockStrategy;
+
+use Prooph\EventStore\Pdo\WriteLockStrategy;
+
+final class MysqlMetadataLockStrategy implements WriteLockStrategy
```
prolic: Not sure if I like the naming, metadata lock makes me think about event metadata. Why did you bring the term "metadata" here? Any reason behind it?

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
fritz-gerneth: This is the MariaDB version of MySQL's `-1`, which MariaDB does not support. Not that it this is missing feature is documented, just not listed as a possbility on the documentation. Took me a good while to figure out why the original `-1` did not work for MariaDB

src/WriteLockStrategy/MysqlMetadataLockStrategy.php
fritz-gerneth: Because `GET_LOCK` uses Mysql's Metadata Locking System. Hence the name specifiying the techonoly used for locking (e.g. in contrast to using a custom roled one like for projections, which still would have a look in its name)

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
prolic: I see, and 6 month is definitely a "-1" :)

src/WriteLockStrategy/MysqlMetadataLockStrategy.php
prolic: Everyday you can learn something new, thanks.

docs/variants.md
prolic: @fritz-gerneth can you  fix that?

prolic:
I'll leave the PR open over the weekend, as I try to implement the postgres version (and test if it's needed).

src/WriteLockStrategy.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo;
+
+interface WriteLockStrategy
+{
+    public function getLock(string $name): bool;
+
+    public function releaseLock(string $name): bool;
```
bwowk: As the release failure is not dealt with (which I guess makes sense since the lock timeout would take care of adverse scenarios) I think this could be a void
```suggestion
    public function releaseLock(string $name): void;
```

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return false;
+    }
+
+    public function releaseLock(string $name): bool
+    {
+        $res = $this->connection->query('SELECT RELEASE_LOCK(\'' . $name . '\') as \'release_lock\'');
+
+        if ($res) {
+            $res->fetchAll();
```
bwowk: why do you fetch it?

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
fritz-gerneth: 0 should work as a timeout I think

fritz-gerneth:
@prolic @codeliner any update on this being merged? Might make sense to merge it without Postgres fixes (if required) for now so this gets fixed for those using MySQL (cc @kochen )?

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
bweston92: @fritz-gerneth that statement doesn't go with the exception, you stated in the exception `$timeout must be larger then 0.`

prolic:
I did not forget about this!

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
fritz-gerneth: Fixed the exception message with the latest commit now

fritz-gerneth:
Updated to release the acquired lock even when throwing an exception. Most notably when throwing concurrency exceptions. 
Noticed that as this left our works stuck waiting for the lock (running in seperate processes, not killed when this particular exception happens).

Sevavietl:
@prolic, @codeliner, @fritz-gerneth, sorry for bothering, but are there any updates on this one?I am using

prolic:
postgres advisory lock strategy implemented via e4fb0e1278ac583b05a52c6126a6f5f402898224

Issue #142, opened 2018-04-09
Select on event stream not using index


elpirata15:
PHP version

- 7.2.2

Prooph packages version

-         "prooph/event-sourcing": "^5.0",
-         "prooph/event-store-symfony-bundle": "dev-master",
-         "prooph/event-store": "^7.0",
-         "prooph/pdo-event-store": "^1.0",
-         "prooph/service-bus-symfony-bundle": "^0.6.0",

Table information

Table: _4228e4a00331b5d5e751db0481828e22a2c3c8e

```
Create Table: CREATE TABLE `_4228e4a00331b5d5e751db0481828e22a2c3c8ef` (
  `no` bigint(20) NOT NULL AUTO_INCREMENT,
  `event_id` char(36) COLLATE utf8_bin NOT NULL,
  `event_name` varchar(100) COLLATE utf8_bin NOT NULL,
  `payload` longtext COLLATE utf8_bin NOT NULL,
  `metadata` longtext COLLATE utf8_bin NOT NULL,
  `created_at` datetime(6) NOT NULL,
  `aggregate_version` int(11) unsigned GENERATED ALWAYS AS (json_extract(`metadata`,'$._aggregate_version')) STORED,
  `aggregate_id` char(36) COLLATE utf8_bin GENERATED ALWAYS AS (json_unquote(json_extract(`metadata`,'$._aggregate_id'))) STORED,
  `aggregate_type` varchar(150) COLLATE utf8_bin GENERATED ALWAYS AS (json_unquote(json_extract(`metadata`,'$._aggregate_type'))) STORED,
  PRIMARY KEY (`no`),
  UNIQUE KEY `ix_event_id` (`event_id`),
  UNIQUE KEY `ix_unique_event` (`aggregate_type`,`aggregate_id`,`aggregate_version`),
  KEY `ix_query_aggregate` (`aggregate_type`,`aggregate_id`,`no`),
  KEY `idx_no` (`no`)
) ENGINE=InnoDB AUTO_INCREMENT=10566693 DEFAULT CHARSET=utf8 COLLATE=utf8_bin
```

show index from _4228e4a00331b5d5e751db0481828e22a2c3c8ef; 

```
+-------------------------------------------+------------+--------------------+--------------+-------------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
[...]

PR #66, opened 2017-03-11
Improve event store indexes

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/43

/cc @basz @oqq

Issue #43, opened 2017-01-17
Improve event store indexes


basz:
referring https://github.com/prooph/pdo-event-store/issues/42 

some indexed need to be changed (ix_unique_event?)

> Than I think ix_unique_event must be reordered
> Or the index must be set on the json path instead..

@oqq

prolic:
@basz Can you verify the problems with the newest master?

basz:
I still seem to have lowering throughput after a while... I don't understand why exactly... The thing is that replaying events has stable performance. That tells me it isn't a problem with the read models (v6 style).

By the way I don't see any difference in the schema, but that might be because I update regularly to master...

oqq:
@basz can you provide a small example gist?
That could me help a lot to improve the indexes.

basz:
would this help? two aggregates one with ~60000 events one with just 1.

https://gist.github.com/basz/f1b6e6c7262e8fabacacb1c05cb35a70

oqq:
Are you using the current implementation from `MySqlSingleStreamStrategy`?

If so please test this query:
```sql
SELECT * FROM _7794a224f85453058cf637d52e7c4487568a8271
    -> WHERE metadata->"$._aggregate_version" > 0 AND metadata->"$._aggregate_id" = '503113e6-820f-5523-b18f-e71a56db31eb' AND metadata->"$._aggregate_type" = 'HF\\Model\\Domain\\Dossier\\Aggregate\\Dossier' AND `no` >= 1
    -> ORDER BY `no` ASC
```

@prolic the main reason for that issue is the lack of indexes.
Maybe it makes sense that we are always provide an index /aggregate_id, aggregate_type, version/ (or in other order) and sort the metadata query always in that order, since this would the most used metadata query I guess. The `MySqlSingleStreamStrategy` stays on unique key, all other could simply use a not unique index.

This would improve performance for such queries against tables with much data.
Currently no index is used at all.

prolic:
I created a test-script for this and am testing now.

oqq:
Please provide `Explain` outputs with used indexes and queries for your tests.

prolic:
I have something else to do now, will come back to it later this evening.

prolic:
Here is the reordered MySqlSingleStreamStrategy: https://gist.github.com/prolic/389820fcd30c821fcce5da3e1e9c6f89

Here are the test-scripts:
[...]

PR #191, opened 2019-03-19
Remove forced index from MySqlSingleStreamStrategy

danizord:
When loading the event stream from a projector, it currently forces MySQL to use the index `ix_query_aggregate`, which contains 3 fields, but in this case the primary key index is more efficient (see https://github.com/prooph/pdo-event-store/issues/142#issuecomment-380023812).

This is causing the DB to run an expensive "Creating sorting index" operation on every query:

image

The change introduced by this PR reduced query time here by 99.7% (from 0.19646025 to 0.00045325 seconds) and reduced the overall CPU usage with 9 projections from 70% to 5% :P

image

Not sure what is the impact when loading events with `MetadataMatcher`, since I'm not reconstituting aggregates from history in my current project. But I guess it should automatically detect the most suited index. Anyway, here is the PR as suggested by @codeliner in gitter.

### DIFF
--- a/src/PersistenceStrategy/MySqlSingleStreamStrategy.php
+++ b/src/PersistenceStrategy/MySqlSingleStreamStrategy.php
@@ -16,12 +16,11 @@ namespace Prooph\EventStore\Pdo\PersistenceStrategy;
 use Iterator;
 use Prooph\Common\Messaging\MessageConverter;
 use Prooph\EventStore\Pdo\DefaultMessageConverter;
-use Prooph\EventStore\Pdo\HasQueryHint;
 use Prooph\EventStore\Pdo\PersistenceStrategy;
 use Prooph\EventStore\Pdo\Util\Json;
 use Prooph\EventStore\StreamName;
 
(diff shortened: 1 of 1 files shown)


prolic:
@danizord The problem is that MySQL is NOT smart enough to figure out the right index. That's really unfortunate, Postgres can do it properly.

See also https://github.com/prooph/pdo-event-store/issues/43 and https://github.com/prooph/pdo-event-store/pull/66.

I know, now the questions comes "But I want to project some event streams and this index is really bad" - and the other side screams "But my event replay is super slow when doing event sourcing". The answer to this is, you obviously don't use the same Persistence Strategy in both cases.

For projections the `SimpleStreamStrategy` works fine, but for event sourcing you need to use the `SingleStreamStrategy`.

I know this whole persistence strategy thingy is not ideal, updated and added already the docs for this multiple times and still people run into issues with this. In v8 things will get different and the persistence strategies will be gone. Until then, use the proper persistence strategy for your use case or create your custom ones.

PR #192, opened 2019-04-01
Fix reference error in PostgresEventStore for MetadataMatcher with IN Operations

fjogeleit:
Fix reference error in PostgresEventStore for MetadataMatcher with IN Operations.

The removed value reference leads to wrong values for the IN operation.

Example:

```
(new MetadataMatcher())
	->withMetadataMatch(
	    'event_name',
	    Operator::IN(),
	    [
	        'Event-1',
	        'Event-2'
	    ],
	    FieldType::MESSAGE_PROPERTY()
	)
```

Results to:

```
$values = [
	'metadata_0_0' => 'Event-1',
	'metadata_0_1' => 'Event-1'
]
```

The Second values is replaced by the value reference with the first one.

### DIFF
--- a/src/PostgresEventStore.php
+++ b/src/PostgresEventStore.php
@@ -688,9 +688,9 @@ SQL;
             $parameters = [];
 
             if (\is_array($value)) {
-                foreach ($value as $k => &$v) {
+                foreach ($value as $k => $v) {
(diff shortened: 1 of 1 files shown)


prolic:
Thanks a lot. I'll try to get to it asap. On the nove right now

PR #202, opened 2019-05-13
Improvement/merged stream iterator

basz:
addresses https://github.com/prooph/pdo-event-store/issues/201
accompanies https://github.com/prooph/event-store/pull/367

Do not merge until ^^ has been released (composer uses https://github.com/basz)

PR #205, opened 2019-05-17
Improvement/persist after blocksize processed events

basz:
accompanies https://github.com/prooph/event-store/pull/368

what PR does is 

Persists the stream position when the number of *processed* events equals the blockSize threshold, not after *handled* events.

Potential harmful for process managers that handle a few event types in a stream. Currently a even projection with blockSize set to 1, it only records it's stream position after a handled events is processed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,15 +16,9 @@
     ],
     "minimum-stability": "dev",
     "prefer-stable": true,
-    "repositories": [
-        {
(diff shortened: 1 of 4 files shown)


basz:
please trigger a rebuild (event-store has changed)

basz:
merge #202. it includes this PR

PR #206, opened 2019-05-26
Fixes projection lock string generation

### LINKED ISSUE #198: Projection fails sometimes when lock timeout is less than 100ms
https://github.com/prooph/pdo-event-store/blob/01b2cf3d30c1d01a7ad87cc00eb968de97c9c79a/src/Projection/PdoEventStoreProjector.php#L991-L1001

sometimes, on the start of a second, the value in microseconds plus the timeout will have less than 6 digits (when the timeout alone doesn't have at least 100000 microseconds), then `\substr($micros, 0, -6)` will return false and modify will fail:

```
DateTimeImmutable::modify(): Failed to parse time string (+ seconds) at position 0 (+): Unexpected character
```

brunowowk:
closes #198

Since PHP 7.1 it's possible to manipulate DateTime instances microseconds, so
I was able to drop the old mechanism that did that manually

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -991,17 +991,8 @@ EOT;
 
     private function createLockUntilString(DateTimeImmutable $from): string
     {
-        $micros = (string) ((int) $from->format('u') + ($this->lockTimeoutMs * 1000));
-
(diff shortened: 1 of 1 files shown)


prolic:
Thanks a lot

PR #207, opened 2019-05-31
Change order of select and count statements to fix concurrency error

unhappyby:
When we make read and write operations with event store at the same time we can face one problem. Making select statement before count can bring us to situation where select statement return, for example, 10 rows and count statement, which will execute after select, return 15 records. It means that we change offset to 15 and lose 5 events.
So I think that we can change order of executing for this two statements and always have count less than select actual row count. This will help us go through as many lines as we get from the count statement. Diff in rows between count and select we can get from next request.

### DIFF
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -353,8 +353,8 @@ EOT;
         }
 
         try {
-            $selectStatement->execute();
             $countStatement->execute();
+            $selectStatement->execute();
         } catch (PDOException $exception) {
             // ignore and check error code
         }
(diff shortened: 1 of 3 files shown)


unhappyby:
Its hard to reproduce in test. Because we need to read and write to event store at the same time in different processes. I think that this changes cannot break anything but they can fix this bug. If you have some questions to understand the problem feel free to ask.

unhappyby:
Can I get any updates for this request, please? Also it would be cool if someone restart travis for passing all checks.

codeliner:
Merged #208 

@unhappyby As stated in the other PR your problem should be solved. Can you verify please?

PR #208, opened 2019-06-02
don't execute count statement on load

prolic:
superseeds https://github.com/prooph/pdo-event-store/pull/207

/cc @basz @fritz-gerneth @codeliner @unhappyby

The tests still pass, so this should work without making any BC. Please let me know what you think.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^7.1",
+        "ext-pdo": "*",
         "prooph/event-store": "^7.5.3"
     },
     "require-dev": {
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -354,7 +354,6 @@ EOT;
 
         try {
             $selectStatement->execute();
-            $countStatement->execute();
         } catch (PDOException $exception) {
             // ignore and check error code
         }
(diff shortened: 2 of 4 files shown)


unhappyby:
I think this is not solving the problem which I describe in [this pull request](https://github.com/prooph/pdo-event-store/pull/207), maybe we need to remove execution in iterator, but not in the event store.

basz:
Please explain why you believe this would not solve the issue.

The PdoStreamIterator does not call count while iterating. So something like the following simply works.

```
/** @var EventStore $eventStore */
$eventStore = $container->get(EventStore::class);
$streamIterator = $eventStore->load(new StreamName('stream_name'), 500);

foreach ($streamIterator as $event) {
  // $event 500 to end (even while added after start)
}
```

The removed code in this PR is here *only* to be able to return an EmptyIterator (useless in my opinion anyway, not sure why it exists...).

unhappyby:
Yes, this code working. But when I use event store in two or more processes (for writing and reading in parallel) some times event store lose events for projections. For example:
For 20000 events event store lose from 1 to 70 events in my project.
That happening because in db, table projections, position sets from count, which I get from `$countStatement` in `load` method. So I think that this PR can make things worse because event store will execute `$countStatement` in `PdoStreamIterator` and this means that gap in time between select and count statements will be bigger.

basz:
I understand the issue but am not sure your problem  comes from the PdoStreamIterator as that does not use the $countStatement unless you specifically call count() on it. In a normal iteration (foreach) it is not used.

basz:
Do you use a loop (with count) in a different usecase? (Have code?)

prolic:
@unhappyby Did you check this PR indepth? As @basz pointed out, the projections don't call `count()`. This PR should indeed resolve your issue (and also give slightly more performance for querying).

codeliner:
Going to merge this one and make a new bugfix release together with #210

PR #210, opened 2019-09-09
Add Schema-Creation to createSchema function

heiglandreas:
Currently when a tableName with a dot is presented the name can correctly
be split up into schema and table name. But when the schema does not
exist the whole proces breaks as the table can not be created in a
non-existing space.

This commit adds an SQL-statement to the createSchema-function that – in
case a schema was given – will create the schema when it does not yet
exist.

### DIFF
--- a/src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
+++ b/src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
@@ -56,9 +56,9 @@ CREATE TABLE $tableName (
 );
 EOT;
 
-        return [
+        return array_merge($this->getSchemaCreationSchema($tableName), [
             $statement,
-        ];
+        ]);
     }
 
     public function columnNames(): array
@@ -99,5 +99,17 @@ EOT;
         }
 
         return $table;
+    }
+
+    private function getSchemaCreationSchema(string $tableName) : array
+    {
+        if (!$schemaName = $this->extractSchema($tableName)) {
+            return [];
+        }
+
+        return [sprintf(
+            'CREATE SCHEMA IF NOT EXISTS %s',
+            $schemaName
+        )];
     }
 }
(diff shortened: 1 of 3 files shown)


codeliner:
I'm going to check why some jobs are failing on travis. Looks like a problem not related to the PR.

codeliner:
Unfortunately, we have an upstream problem that breaks our tests.

Two issues need to be resolved first:
- [ ] https://github.com/prooph/event-store/pull/378
- [ ] https://github.com/prooph/pdo-event-store/pull/213

@prolic We could also merge this PR, b/c failing tests are not related to changes made here. What do you think?

src/PersistenceStrategy/PostgresSimpleStreamStrategy.php
```diff
@@ -100,4 +100,16 @@ public function generateTableName(StreamName $streamName): string
 
         return $table;
     }
+
+    private function getSchemaCreationSchema(string $tableName) : array
```
prolic: `getSchemaCreationSchema` sounds a bit off with those 2 schema in the name

heiglandreas: Sounds off. But according to the current naming scheme the SQL-statement that is sent to the database is called a scheme. And the structure that will be created in the database by this scheme is ... well... a scheme. So either all methods handling/creating SQL that shall be executed would need to be renamed or Postgres needs to rename their database-structure called schema. 

I could of course rename the function to `createDatabaseSchemaCreationSchema`. Not sure though that's better....

Is that a problem that prohibits merging the PR?

prolic: @heiglandreas Thanks for the explanation. All good, we'll merge this PR soon. Just waiting for some feedback on the other issue.

codeliner:
I'm merging the PR. Upstream issues are resolved.

PR #211, opened 2019-09-09
Require PHP 7.2

codeliner:
Event Store requires PHP 7.2 and travis fails for PHP 7.1 due to
a dependency extended by event-store.

This should solve failing tests in #210 caused by latest version of webmozart/assert extended by prooph/event-store under PHP 7.1


codeliner:
Closing this one. Requiring PHP 7.2 is the wrong solution. I'll submit another PR with the pipeline fix only.

PR #212, opened 2019-09-09
Try setting trusty distro on a few tests.

bweston92:


codeliner:
I've set distro to bionic in my PR: https://github.com/prooph/pdo-event-store/pull/211/files#diff-354f30a63fb0907d4ad57269548329e3R9

Travis takes some time at them moment, but first tests are green (including MySql). Looks like this will solve the issue.

PR #213, opened 2019-09-09
Use bionic distro and mysql as service

codeliner:


codeliner:
Tests will fail as long as we don't have a solution for this: https://github.com/prooph/event-store/pull/378

prolic:
thanks

PR #217, opened 2019-12-09
Prevents using incompatible persistence strategy

gquemener:
Currently, configuring a MariaDB PersistenceStrategy within a PostgresEventStore will not trigger PHP type check.
The result will most likely be the db server complaining about a wrong syntax (in the best case).

This PR prevents mixing persistence strategies and event store implementations.

Note: Not sure if this is compatible with php <7.4

### DIFF
--- a/src/MariaDbEventStore.php
+++ b/src/MariaDbEventStore.php
@@ -25,6 +25,7 @@ use Prooph\EventStore\Metadata\Operator;
 use Prooph\EventStore\Pdo\Exception\ConcurrencyExceptionFactory;
 use Prooph\EventStore\Pdo\Exception\ExtensionNotLoaded;
 use Prooph\EventStore\Pdo\Exception\RuntimeException;
+use Prooph\EventStore\Pdo\PersistenceStrategy\MariaDbPersistenceStrategy;
 use Prooph\EventStore\Pdo\Util\Json;
 use Prooph\EventStore\Pdo\WriteLockStrategy\NoLockStrategy;
 use Prooph\EventStore\Stream;
@@ -45,7 +46,7 @@ final class MariaDbEventStore implements PdoEventStore
     private $connection;
 
     /**
-     * @var PersistenceStrategy
+     * @var MariaDbPersistenceStrategy
      */
     private $persistenceStrategy;
 
@@ -80,7 +81,7 @@ final class MariaDbEventStore implements PdoEventStore
     public function __construct(
         MessageFactory $messageFactory,
         PDO $connection,
-        PersistenceStrategy $persistenceStrategy,
+        MariaDbPersistenceStrategy $persistenceStrategy,
         int $loadBatchSize = 10000,
         string $eventStreamsTable = 'event_streams',
         bool $disableTransactionHandling = false,
--- a/src/MySqlEventStore.php
+++ b/src/MySqlEventStore.php
@@ -25,6 +25,7 @@ use Prooph\EventStore\Metadata\Operator;
 use Prooph\EventStore\Pdo\Exception\ConcurrencyExceptionFactory;
 use Prooph\EventStore\Pdo\Exception\ExtensionNotLoaded;
 use Prooph\EventStore\Pdo\Exception\RuntimeException;
+use Prooph\EventStore\Pdo\PersistenceStrategy\MySqlPersistenceStrategy;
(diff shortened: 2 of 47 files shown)


prolic:
Nice, that's a good improvement, a lot of people made exactly that mistake.

fritz-gerneth:
> Note: Not sure if this is compatible with php <7.4

From a quick glance I have not seen anything incompatible. Any specific concerns?

gquemener:
> > Note: Not sure if this is compatible with php <7.4
> 
> From a quick glance I have not seen anything incompatible. Any specific concerns?

I was thinking about that https://www.php.net/manual/en/migration74.new-features.php#migration74.new-features.core.type-variance, but it's not related to the content of this PR.

fritz-gerneth:
Since this has not been released yet, I only noticed this now while testing #221 : The constructor signature change from the generic `PersistenceStrategy` to a vendor specific interface itself is a BC break for those using custom persistence strategies (like us :)). Not a big deal and easy to be fixed, but something that should be announced respectively.

prolic:
Constructor is not part of the interface

gquemener:
> Constructor is not part of the interface

Agreed. However could you be more specific and point to some code (or add a snippet to your comment), please? I'd like to fully understand what's the issue here.

In case you missed it, all the 3 new interfaces extends the existing `PersistenceStrategy` interface, which will allow you to keep the typehint of your dependenc(ies) with this one, thus preventing a BC break.

fritz-gerneth:
Yes, that's the issue I found myself in. Not exactly sure how many people use a custom strategy and the change is very small. Strictly folllowing semver would require a new major version indeed. 
Personally both approaches (keeping, doing the check in the constructor) are fine. My initial concern only was that this should not go unmentioned and at the least be mentioned in the release notes somewhere :)

codeliner:
Puh, I missed the discussion (hadn't much time for OSS end of last year). I was pointed to the BC break introduced in this PR today. event-store v7 is downloaded 400 times / day.

I really don't like to publish any BC breaks even not small ones, when a library is used heavily in production environments (which is the case for prooph event store v7).

We should rollback the changes. They are a good thing to prevent stupid use of persistence strategies. That's fine. But only a major release would allow us to make life better for new users. At the moment we should focus more on existing users, because we have a lot!

codeliner:
What we could do is, we start working on 2.0, but gap detection needs to be included in 1.x 

My suggestion:
- PR to rollback changes
- finish gap detection
- make a new 1.12 release
- make 1.x branch 
- use master branch for 2.0 release preparation and make PR against master again

gquemener:
> PR to rollback changes
I can take care of it. WDYT of introducing the interfaces in a BC manner (aka reverting only the event stores ctor signature and adding extra type check)?

gquemener:
Maybe even [trigger a deprecate error](https://www.php.net/manual/fr/function.trigger-error.php) when a non vendor specific persistence strategy is injected :thinking:

codeliner:
Yes, deprecation error please, but not an exception, otherwise you still have the problem that your  custom strategy will no longer work without implementing the new interface

codeliner:
and thank you @gquemener for the PR and taking care of it :+1:

prolic:
Looks all good to me, only one thing: I'm a little hesitant starting a 2.0 branch and release, when all we have as changes is this persistence strategy thing. For a new major release I would expect a little more than only "cosmetic" changes.

prolic:
> Constructor is not part of the interface

I was writing from phone and didn't realize this in indeed a BC. Sorry for the confusion.

fritz-gerneth:
Deprecation error is a good idea 👍

codeliner:
@prolic release fast, release often :D

but yeah, we don't need to release 2.0 soon. The added deprecation warning should already be a good hint for those who use persistence strategy wrong. 

But maybe we already have some more topics to tackle in a 2.0 like gap detection enabled by default along with a rewrite of the projections to  decrease constructor arguments and better share functionality between query, projection and read model projection.

I'd also like to work on Postgres Notify Support. Not sure if this can be added without BC break to 1.x

prolic:
Whoever wants to work on those topics, feel free to do so. I can do code
reviews but won't have much time coding this.

I'll prepare 1.12 release this weekend.

@codeliner anything else you need to do before this can get merged?

Issue #189, opened 2019-01-30
MySQL Projections skipping events (SingleStreamStrategy)


fritz-gerneth:
**Issue**: under unknown circumstances the `PdoEventStoreReadModelProjector` skips applying events to registered handlers. 

I have seen this issue happen a few times over the past year only. Yet since missing events in the projections causes quite a few problems I'm trying to figure out the issue since then. Creating this issue for tracking & by any chance input from others. 

**Setup**: Mysql EventStore & Projections with SingleStreamStrategy running in their own process. DB Server and process server running on different VM hosts.

Projection options:
```
PdoEventStoreReadModelProjector::OPTION_PERSIST_BLOCK_SIZE => 1,
PdoEventStoreReadModelProjector::OPTION_PCNTL_DISPATCH => true,
PdoEventStoreReadModelProjector::OPTION_LOCK_TIMEOUT_MS => 20000,
PdoEventStoreReadModelProjector::OPTION_UPDATE_LOCK_THRESHOLD => 15000,
```

**Symptoms**: take a simple set of `IssueCreated`, `IssueUpdated`, `IssueDeleted` of any examplary `Issue` AR. Events occuring in the typical order of `IssueCreated` -> `IssueUpdated` (n times) -> `IssueDeleted`. The projection simply maintains a list of all (undeleted) issues and their latest values. Each event has a handler registered to either insert a row, update a row or delete the row respectively. 
For 99.9% of the time this works perfectly fine as expected. Yet for unknown reasons very rarely some event is not handled. The effect on the projection differes depending on the event (e.g. `IssueUpdate` -> update lost) but is particular bad for `IssueCreated` as rows are missing at all (subsequent events might rely on the presence). 
Resetting the projection solves this issue and all events are applied as expected. But this is a pretty unpractical to do in a production environment when this takes days each time. 

**Debugging**:
To help me make this problem visible in the first place I slightly modified the `[PdoEventStoreReadModelProjector](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreReadModelProjector.php#L575)` to simply log a line about which event it is now dispatching:

```
foreach ($events as $key => $event) {
	$this->logger->info(sprintf(
		'Dispatching event %s[%s]',
		$streamName,
		$key
	));

	if ($this->triggerPcntlSignalDispatch) {
		\pcntl_signal_dispatch();
	}
```

In normal operations I get sequential messages in this format ( `gcl_collection` is my stream's name):
```
[...]

PR #221, opened 2020-02-22
Add gap detection to projections

codeliner:
Fixes https://github.com/prooph/pdo-event-store/issues/189

The PR adds a "gap detection" mechanism to projections. This means, that if the projection detects a gap in the stream position while processing events, it pauses processing for a configurable time, before rereading events from the last known position. 
Gap detection is disabled by default.

## Enabling gap detection for a projection:

```php
$gapDetection = new GapDetection();

$projection = $projectionManager->createProjection('test_projection', [
            PdoEventStoreProjector::OPTION_GAP_DETECTION => $gapDetection,
        ]);
``` 

## Enabling gap detection for a read model projection with custom configuration:

```php
$gapDetection = new GapDetection(
    // Configure retries in case a gap is detected
    [
        0, // First retry without any sleep time
        10, // Second retry after 10 ms
        30, // Wait another 30 ms before performing a third retry
    ],
    \DateInterval('PT60S') //Set detection window to 60s
);

$projection = $projectionManager->createReadModelProjection('test_projection', [
            PdoEventStoreReadModelProjector::OPTION_GAP_DETECTION => $gapDetection,
        ]);
``` 

### Detection window
If you set a detection window, gap detection is only performed on events not older than `NOW - window`
If the detection window is set to `PT60S`, only events created within the last 60 seconds are checked. Be careful when setting `Event::createdAt` manually (for example during a migration).

### DIFF
--- a/src/Projection/GapDetection.php
+++ b/src/Projection/GapDetection.php
@@ -37,7 +37,8 @@ final class GapDetection
     private $retryConfig = [
         0, // First retry is triggered immediately
         5, // Second retry is triggered after 5ms
-        50, // Either DB is really busy or we have a real gap, wait another 50ms and run a last try
+        50, // Third retry with much longer sleep time
+        500, // Either DB is really busy or we have a real gap, wait another 500ms and run a last try
         // Add more ms values if projection should perform more retries
     ];
 
--- a/src/Projection/GapDetection.php
+++ b/src/Projection/GapDetection.php
@@ -1,0 +1,114 @@
+<?php
+
+/**
+ * This file is part of prooph/pdo-event-store.
+ * (c) 2016-2020 Alexander Miertsch <kontakt@codeliner.ws>
+ * (c) 2016-2020 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Pdo\Projection;
+
+use Prooph\Common\Messaging\Message;
+
+/**
+ * Class GapDetection
+ *
+ * In high load scenarios it is possible that projections skip events due to newer events becoming visible before old ones.
+ * For details you can take a look at this issue: https://github.com/prooph/pdo-event-store/issues/189
+ *
+ * The GapDetection class helps projectors to detect gaps while processing events and coordinates retries to fill gaps.
+ *
+ * @package Prooph\EventStore\Pdo\Projection
+ */
+final class GapDetection
+{
+    /**
+     * If gap still exists after all retries, projector moves on
+     *
+     * You can set your own sleep times. Each number is the sleep time in milliseconds before the next retry is performed
+     *
+     * @var array
+     */
+    private $retryConfig = [
+        0, // First retry is triggered immediately
+        5, // Second retry is triggered after 5ms
+        50, // Either DB is really busy or we have a real gap, wait another 50ms and run a last try
+        // Add more ms values if projection should perform more retries
+    ];
+
+    /**
+     * There are two reasons for gaps in event streams:
+     *
+     * 1. A transaction rollback caused a gap
+     * 2. Transaction visibility problem described in https://github.com/prooph/pdo-event-store/issues/189
+     *
+     * The latter can only occur if a projection processes events near realtime.
+     * The detection window ensures that during a projection replay gap retries are not performed. During replays
+     * only gaps caused by transaction rollbacks are possible. Retries would just waste a lot of time and resources.
+     *
+     * By default the window is set to 30 seconds, meaning only if the current processed event was recorded in the last
+     * 30 seconds gap detection will tell the projector to perform a retry.
+     *
+     * @var \DateInterval
+     */
+    private $detectionWindow;
+
+    /**
+     * @var int
+     */
+    private $retries = 0;
+
+    public function __construct(array $retryConfig = null, \DateInterval $detectionWindow = null)
+    {
+        if ($retryConfig) {
+            $this->retryConfig = $retryConfig;
+        }
+
+        if ($detectionWindow) {
+            $this->detectionWindow = $detectionWindow;
+        } else {
+            $this->detectionWindow = new \DateInterval('PT30S');
+        }
+    }
+
+    public function isRetrying(): bool
+    {
+        return $this->retries > 0;
+    }
+
+    public function trackRetry(): void
+    {
+        $this->retries++;
+    }
+
+    public function resetRetries(): void
+    {
+        $this->retries = 0;
+    }
+
+    public function getSleepForNextRetry(): int
+    {
+        return (int) $this->retryConfig[$this->retries] ?? 0;
+    }
+
+    public function isGapInStreamPosition(int $streamPosition, int $eventPosition): bool
+    {
+        return $streamPosition + 1 !== $eventPosition;
+    }
+
+    public function shouldRetryToFillGap(\DateTimeImmutable $now, Message $currentMessage): bool
+    {
+        //Only retry if message was written to stream in last 30s, otherwise it's a gap caused by a transaction rollback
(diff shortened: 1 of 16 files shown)


codeliner:
Some more tests + documentation is missing, but I'd like to collect some feedback on the PR. We will run some experiments next week, to check if gap detection solves the issue of skipped events.

src/Projection/GapDetection.php
```diff
@@ -0,0 +1,106 @@
...
+ * Class GapDetection
+ *
+ * In high load scenarios it is possible that projections skip events due to newer events becoming visible before old ones.
+ * For details you can take a look at this issue: https://github.com/prooph/pdo-event-store/issues/189
+ *
+ * The GapDetection class helps projectors to detect gaps while processing events and coordinates retries to fill gaps.
+ *
+ * @package Prooph\EventStore\Pdo\Projection
+ */
+final class GapDetection
```
prolic: We need an interface so we can test against mocks.

src/Projection/PdoEventStoreProjector.php
```diff
@@ -169,6 +171,11 @@ final class PdoEventStoreProjector implements Projector
      */
     private $metadataMatcher;
 
+    /**
+     * @var GapDetection/null
```
prolic: `|null` not `/null`

src/Projection/PdoEventStoreProjector.php
```diff
@@ -180,7 +187,8 @@ public function __construct(
         int $persistBlockSize,
         int $sleep,
         bool $triggerPcntlSignalDispatch = false,
-        int $updateLockThreshold = 0
+        int $updateLockThreshold = 0,
+        GapDetection $gapDetection = null
```
prolic: I really like how the amount of constructor arguments grow and grows... way beyond anything initially designed. But that's not your fault here, just something I noticed.

src/Projection/GapDetection.php
```diff
@@ -0,0 +1,106 @@
...
+    public function __construct(array $retryConfig = null, \DateInterval $detectionWindow = null)
+    {
+        if ($retryConfig) {
+            $this->retryConfig = $retryConfig;
+        }
+
+        $this->detectionWindow = $detectionWindow;
+    }
+
+    public function isRetrying(): bool
```
prolic: Without having a direct advice here: I feel like this should be easier to use from the outside. Something along the lines of:
```
$gapDetector->handlePossibleGaps($arg1, $arg2);
```

while the rest is hidden behind implementation details. Right now it forces me to know some details and call methods in a particular order. On the other hand, it's only called from within the lib and not from userland code, so maybe I'm a little too critical here.

prolic:
I'm not sure about those predefined timing windows. Some transactions might be long-running and the sleep times are fairly short. From what I saw in the past, those gabs are pretty rare, so even if the projection falls behind for 5 secs, that's still acceptable in most situations, so maybe we should add some longer retry times as well. But without data that's only speculation.

src/Projection/PdoEventStoreProjector.php
codeliner: yeah, one reason why I've put the gap detection stuff into a class and use it as config option. otherwise we would have 3 more constructor args ;)

src/Projection/GapDetection.php
codeliner: I don't think the class needs to be mocked. It is basically a wrapper around some config options and two helper methods. See current test case. Works without mocking.

src/Projection/GapDetection.php
codeliner: This would require a rewrite of the projection logic. Something I'd like to avoid for v7. Might be a good thing for v8, though.

codeliner:
"normal gaps" can occur very easily. They are caused by concurrency exceptions. That's the reason why I'd like to avoid a long sleep time by default. 50ms are already a very long time for 99% of write operations in a typical event sourced implementation.

fritz-gerneth:
This is pretty awesome stuff and I'll put this to testing on our side this week too!

codeliner:
Cool, looking forward to your results @fritz-gerneth 
We performed first tests yesterday but still had gaps with the default sleep times. So I guess @prolic is right that we need longer defaults. But we're testing Postgres event-store without `GET_LOCK`. 
Hopefully we can perform some more tests tomorrow.

`GET_LOCK` + gap detection should work with short sleep times, right? Because in that case faulty gaps only occur when one aggregate records multiple events with different payload size in one transaction.

fritz-gerneth:
> GET_LOCK + gap detection should work with short sleep times, right? Because in that case faulty gaps only occur when one aggregate records multiple events with different payload size in one transaction.

I think so. 

We ran error-prone operations a couple of thousand times today without any issue (default gap detection settings). Error-prone in the magnitude of "once every 1k - 2k saves" though..

codeliner:
Great news @fritz-gerneth :tada: 

We can also confirm that for our case a longer sleep time fixed the issue (No GET_LOCK, only gap detection). We're trying to balance sleep time vs. gaps now.

So I'll definitely continue with the PR and finish it. Sleep times need to be documented anyway, but I think I'd go with @prolic suggestion and increase the default sleep time. If it is too slow for someone, they can simply override the setting and define their own.

prolic:
I think you can still use the sleep times you have right now, just add more retries with larger intervals. I would be totally fine with a projection hanging 5 secs every now and then, when it means it processes correctly. Way better than handling stuff quick and still having gabs.

ahmed-alaa:
@codeliner I tried the solution on Postgres with max retry of 100ms as well as windows time of 120s and without `GET_LOCK` strategy.

So far It works good without missing events within 10,000 events in average of 15-20 events/second.

Please Note: for example with our current implementation to override the `PostgresSingleStreamStrategy`  or `MySqlSingleStreamStrategy`, this might break any service that does the same because of these changes https://github.com/prooph/pdo-event-store/commit/2b2f943b717487a2a045a168186246102c787b86#diff-38bb66b2f052fb5ca610f76079f40f3f, would be nice if could take that into consideration during the release :) 

Thanks alot for the fix!

codeliner:
thx @ahmed-alaa I started a discussion about BC break for persistence strategies. This will not be included in a 1.12 release: https://github.com/prooph/pdo-event-store/pull/217#issuecomment-592454046

src/Projection/PdoEventStoreProjector.php
```diff
@@ -603,7 +622,7 @@ private function fetchRemoteStatus(): ProjectionStatus
         return ProjectionStatus::byValue($result->status);
     }
 
-    private function handleStreamWithSingleHandler(MergedStreamIterator $events): void
+    private function handleStreamWithSingleHandler(MergedStreamIterator $events): bool
```
gquemener: It personnally would feel more easy to follow execution flow if gap detection would be communicated through an exception, instead of a boolean.

prolic: Exceptions are more costly to build than simply returning a bool value. Also this is a private method, we are not communicating anything to the enduser.

codeliner:
@prolic I guess you mean this PR in your [question](https://github.com/prooph/pdo-event-store/pull/217#issuecomment-592737725), right? 

## Open Todos
- [ ] more tests
- [ ] Documentation

I try to finish it on Sunday.

prolic:
Perfect, thanks

src/Projection/PdoEventStoreProjector.php
codeliner: This :point_up: and in the run method the returned boolean is only checked if gap detection is enabled, otherwise it is completely ignored and has no effect.

codeliner:
@prolic ready to merge

keywinf:
Very nice work, I personally used a Fibonacci sequence (0, 5, 5, 10, 15, 25, ... until 2s) as GapDetection options. Thanks

webdevilopers:
> ## Open Todos
> 
>     * [x]  more tests
> 
>     * [x]  Documentation

I guess the `OPTION_GAP_DETECTION` option should be added here:
* http://docs.getprooph.org/event-store/projections.html

PR #222, opened 2020-02-28
Restores custom strategy backward compatibility

gquemener:
This PR prevents a BC break introduced in https://github.com/prooph/pdo-event-store/pull/217 by reverting the event stores constructor signature.

It also triggers a deprecation error to prepare users to future BC break (in 2.0.0).


codeliner:
Awesome! Thank you @gquemener

codeliner:
@ahmed-alaa can you have a look as well?

ahmed-alaa:
@codeliner I tried it and works as expected 👍

Issue #224, opened 2020-04-14
Running projections causes constant full table scans due incorrect forced index


atymic:
We're developing a pretty new app using prooph, with only ~20k events in a single stream. We have ~30 projections, and when running these all at once the database gets completely destroyed.

I dived in with a debugger and figured out this is because we are doing 30x full table scans per 100ms, as every loop of each projection is doing a full table scan.

Here's an example query:
```sql
SELECT * FROM `_4228e4a00331b5d5e751db0481828e22a2c3c8ef` USE INDEX(ix_query_aggregate)
WHERE `no` >= 19067
ORDER BY `no` ASC
LIMIT 1000;
```

![image](https://user-images.githubusercontent.com/50683531/79175896-417eed00-7e42-11ea-9d7a-28f54ac9fe7b.png)

Note the `USE INDEX(ix_query_aggregate)`. This forces mysql to not use the the primary index on `no`.

Removing the `USE INDEX` means that the index is used properly, and the query is orders of magnitude faster.

![image](https://user-images.githubusercontent.com/50683531/79175903-43e14700-7e42-11ea-9a0e-05e84878a873.png)

Apologies if this is a misconfiguration from our end, but I think that the `ix_query_aggregate` index should only be using when doing queries based on metadata, not when scanning the event stream.

Running:
`prooph/pdo-event-store`: `v1.12.0`
`prooph/event-store`: `v7.5.6`

Thanks in advanced!

prolic:
Can you post your configuration please?

atymic:
<details>
<summary>Config</summary>
```
return [
    /*
    |--------------------------------------------------------------------------
    | Plugins
    |
    | List of all global event store plugin service IDs
    |
    |--------------------------------------------------------------------------
    */
    'plugins' => [],

    /*
    |--------------------------------------------------------------------------
    | Metadata Enrichers
    |
    | List of all global event store metadata enrichers
    |
    |--------------------------------------------------------------------------
    */
    'metadata_enrichers' => [],

    /*
    |--------------------------------------------------------------------------
    | Event Stores
    |--------------------------------------------------------------------------
    |
    | Each event store will be configured here. Currently only the mysql,
    | maria_db and postgres stores are supported. Each store will be bound to
    | event_store.stores.<key> as well as to the EventStore FQCN. Available
    | settings are:
    |
    | - persistence_strategy: The class name or service ID of the persistence strategy
[...]

PR #225, opened 2020-04-15
Add indexed queries to MySqlEventStore

marcuswinkler:
We recently noticed that the MySql event store does not actually use the index when loading events for an aggregate.

This PR copies the implementation from the Maria DB event store. Note that it does so only for the single stream strategy, as this is what we use.

We have been running this in production for several weeks now (patched on top of 1.10.5).

### DIFF
--- a/src/MySqlEventStore.php
+++ b/src/MySqlEventStore.php
@@ -721,6 +721,7 @@ SQL;
         }
 
         foreach ($metadataMatcher->data() as $key => $match) {
+            $this->convertToColumn($match);
             /** @var FieldType $fieldType */
             $fieldType = $match['fieldType'];
             $field = $match['field'];
@@ -863,6 +864,22 @@ EOT;
 
             if (! $result) {
                 throw new RuntimeException('Error during createSchemaFor: ' . \implode('; ', $statement->errorInfo()));
+            }
+        }
+    }
+
+    /**
+     * Convert metadata fields into indexed columns
+     *
+     * @example `_aggregate__id` => `aggregate_id`
+     */
+    private function convertToColumn(array &$match): void
+    {
+        if ($this->persistenceStrategy instanceof MySqlIndexedPersistenceStrategy) {
+            $indexedColumns = $this->persistenceStrategy->indexedMetadataFields();
+            if (\in_array($match['field'], \array_keys($indexedColumns), true)) {
+                $match['field'] = $indexedColumns[$match['field']];
+                $match['fieldType'] = FieldType::MESSAGE_PROPERTY();
             }
         }
     }
--- a/src/MySqlIndexedPersistenceStrategy.php
+++ b/src/MySqlIndexedPersistenceStrategy.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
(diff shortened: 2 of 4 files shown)


codeliner:
Thx for the PR @marcuswinkler 

Can you elaborate more on the problem? In 2017 @prolic and @ogg did a lot of testing to figure the best index for MySql. It is documented here: https://github.com/prooph/pdo-event-store/issues/43#issuecomment-285877797 along with test scripts etc.

What MySql version are you using and can you provide some `EXPLAIN` results, that show that the index is not used. Also please share your event store configuration that you used for testing, before patching the strategy.

The other issue is, that I don't think we can merge the PR as-is. This would have a huge impact on existing installations and would break existing streams.

So let us first figure out your exact problem and then discuss a solution. We did a similar thing yesterday for MySql having problems with projections: https://github.com/prooph/pdo-event-store/issues/224

marcuswinkler:
Well that's a bit embarrassing.

Trying to prove with `EXPLAIN` that this PR is necessary, I now find that the correct index is used in master after all.

My assumption that MySql would not use the index for generated columns unless the generated column was directly referenced in the query was incorrect. Therefore it is also not required to translate the column names as in the Maria DB event store.

(See also https://dev.mysql.com/doc/refman/5.7/en/generated-column-index-optimizations.html)

A few weeks ago, cpu usage on our database instance was shooting up to 100%. Looking at the queries in the process list, I found that many of the following took minutes to complete:
```
SELECT * FROM `events` USE INDEX(ix_query_aggregate, PRIMARY)
WHERE JSON_UNQUOTE(metadata->"$._aggregate_type") = 'email_identity'  AND JSON_UNQUOTE(metadata->"$._aggregate_id") = '7678e75f-85a9-4c97-b5c7-c12575f3634c'  AND `no` >= 1;
```

When I examined those with `EXPLAIN` they were _not_ using an index. Running it right now on the same database they _do_ use the index:
```

+----+-------------+--------+------------+-------+----------------------------+--------------------+---------+------+------+----------+-----------------------+
| id | select_type | table  | partitions | type  | possible_keys              | key                | key_len | ref  | rows | filtered | Extra                 |
+----+-------------+--------+------------+-------+----------------------------+--------------------+---------+------+------+----------+-----------------------+
|  1 | SIMPLE      | events | NULL       | range | PRIMARY,ix_query_aggregate | ix_query_aggregate | 754     | NULL | 36 |   100.00 | Using index condition |
```

However, the changes in this PR did immediately fixed our cpu usage issues at the time. I have no explanation for that.

Database: MySql 5.7.17 on an AWS t2.small instance
Using a MySqlSingleStreamStrategy with ~ 5 million events, ca. 4.3 GB in the events table
Typical cpu usage < 10 %

codeliner:
Good to know. Thx for the feedback. Any other result would have been strange, because pdo-event-store is available since 3 years and downloaded ~400 times a day, meaning many teams are using it in production. We also ran a lot of benchmarks and never noticed the lack of a missing or wrongly configured index. But you never know ;)
One thing I could imagine is that you somehow used the `SimpleStreamStrategy` instead of the `SingleStreamStrategy`.

Issue #226, opened 2020-11-17
Aggregate root with `.` in ID will break PosgreSQL PostgresAggregateStreamStrategy


patrickkusebauch:
I was able to drill it down to this line: https://github.com/prooph/pdo-event-store/blob/8f8c3621a8dacd5fc9ee8299e42dc552188b0ea4/src/PersistenceStrategy/PostgresAggregateStreamStrategy.php#L104

## Current behavior/ Steps to reproduce/issue:
 1. Create an aggregate that has `.` in aggregateRootId (example: `patrick.kusebauch`). 
 2. Try to store it with `PostgresAggregateStreamStrategy`
 3. The name of the table before the referenced line: `_ff0a332cf5da03cc35df5dcceb1b87ba21af8c95` (41 chars)
 4. The name of the table after the referenced line: `Full\Path\To\The\Class\Aggregates\User-patrick._ff0a332cf5da03cc35df5dcceb1b87ba21af8c95`
 5. The `event_streams` table's column `stream_name`  is `char(41)`
 6. Try to store the "wrong" table name
 7. `Prooph\EventStore\Pdo\Exception\RuntimeException('Error 22001. Maybe the event streams table is not setup? Error-Info: ERROR:  value too long for type character(41)')`

## Expected behavior
Able to create an aggregate with an ID that contains `.`

prolic:
Can you try to change the function here? https://github.com/prooph/pdo-event-store/blob/master/src/Util/PostgresHelper.php#L44

patrickkusebauch:
@prolic I do not understand what is the purpose of that function in the context of this package, as to me, from the bug description, it seems as this function should no be there at all.

prolic:
What schema are you using in postgres? I would assume if you never used this value it should be default to `public`? At least as a workaround, try this `public.patrick.kusebauch` instead of this `patrick.kusebauch`. In postgres schema and table name are separated by a dot, hence we extract the schema name from the string as everything that is coming before the first dot.

patrickkusebauch:
@prolic sorry for the delay.
I use the Posgres:12 docker container, with no extra settings.
For table definitions, I use the [provided SQL](https://github.com/prooph/pdo-event-store/blob/master/scripts/postgres/01_event_streams_table.sql).

codeliner:
[...]

PR #227, opened 2020-12-14
Failing test

patrickkusebauch:
Failing test for issue #226

### DIFF
--- a/tests/PostgresEventStoreTest.php
+++ b/tests/PostgresEventStoreTest.php
@@ -81,6 +81,20 @@ class PostgresEventStoreTest extends AbstractPdoEventStoreTest
     /**
      * @test
      */
+    public function it_created_aggregate_with_id_containing_dot_char_while_using_aggreagete_stream_Strategy(): void
+    {
+        $this->setupEventStoreWith(new PostgresAggregateStreamStrategy());
+        $streamName = new StreamName('Prooph\Model\User-aggregate.id');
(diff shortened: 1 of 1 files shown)


prolic:
Will look into it asap, super busy these days.

codeliner:
@prolic @fritz-gerneth any ideas what would be the best way to solve the problem?

a) Escape dots in aggregate id by default
b) Document that a custom stream strategy should be used in this case

patrickkusebauch:
@codeliner not directed at me, but I think just documenting it is a bad idea. At least, it should throw an Exception. But then you would need to change to the method signature. That would be a breaking change. Escaping by default would be great.

prolic:
I tried to look into this today. I don't think there is a solution to this problem without BC. Everything before the dot is treated as the schema name (in postgres that's "public" by default) but can be any other schema as well.

I suggest you use another delimiter for separation of aggregate name and id.

@codeliner @fritz-gerneth @enumag thoughts?

codeliner:
A breaking change is no option at this point and I also don't have another idea. One needs to accept this limitation.
Since AggregateStreamStrategy is not the best choice for most use cases because projections are too slow (see note [here](http://docs.getprooph.org/event-store/implementations/pdo_event_store/variants.html#3-2-1-2-5)), I consider this an edge case. I'm not happy with it, but time is limited.

PR #228, opened 2021-01-08
Php8 support

fritz-gerneth:
- [x] Upate travis links
- [x] Update to PHPUnit 9
- [x] Update DocHeader
- [x] Add PHP8 to travis
- [ ] Split PHPCs in travis configuration
- [x] Update to Xdebug3 / fix test coverage 
- [ ] ~~Update `event-store` repository to mark inherited tests as `@long`~~


fritz-gerneth:
>  Update event-store repository to mark inherited tests as @long

This is a mean one. The tests here extend abstract test classes from the `event-store` repository: [`AbstractEventStoreTest`](https://github.com/prooph/event-store/blob/7.x/tests/AbstractEventStoreTest.php) in particular. In the base repository these tests are fast running, in here we need longer test execution. 

To fix this we'd need to 
- Add these tags in the base repo
- Depend on the newley released version (^7.6.1)

In first I was a bit hesitant to just bump the version requirement because of these tags (as for functionality we could still use 7.5.x). But tests in here would not run with 7.5.x at all anyway. So I think the next steps would be to add the tags in the `event-store` repo, release a new version, and depend on this then.

fritz-gerneth:
Tests should all run once we can depend on https://github.com/prooph/event-store/pull/418

netiul:
@fritz-gerneth Perhaps those tests that need to run longer can be overridden in the concrete classes in this library and given the correct annotation. It would make the tests here a bit more robust by not relying on annotations in a different library.

codeliner:
@fritz-gerneth what do you think about the suggestion by @netiul ? Sounds like a good idea to me or is it not possible?

fritz-gerneth:
> 
> 
> @fritz-gerneth what do you think about the suggestion by @netiul ? Sounds like a good idea to me or is it not possible?

I am not exactly a fan of it but might be the best option. Will try this the next days.

fritz-gerneth:
Overwriting does work indeed, though I don't like this solution, or rather more the annotations themselves. They make the tests brittle (e.g. they fail on travis, not my dev environment due to external factors). On the long I'd suggest to disable test timers completely instead. 

Everything else should be completed, with modern phpunit and xdebug3, test coverage & CS work again too.

codeliner (approved):

codeliner:
Thank you so much for taking care of this @fritz-gerneth :clap:

PR #234, opened 2021-11-10
free up memory from PDO statement before next run

sandrokeil:
This fixes a memory issue if you have large events or a large `loadBatchSize`. 

Assume you have a Docker service memory limit of 512 MB. The first run (`$selectStatement->execute();` in `\Prooph\EventStore\Pdo\PostgresEventStore::load`) consumes 320 MB of RAM because of big events and a large load batch size. This is stored in the `$streamEvents` variable.

On the second run, another bunch of events with 320 MB of RAM will be loaded and freeing up the `$streamEvents` variable is to late. That's why it should be cleaned up at the beginning of the each run.

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -522,6 +522,7 @@ EOT;
         try {
             do {
                 $eventStreams = [];
+                $streamEvents = []; // free up memory from PDO statement
 
                 foreach ($this->streamPositions as $streamName => $position) {
                     try {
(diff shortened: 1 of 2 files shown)


kochen:
Hey @sandrokeil nice catch!
Would it be possible for you to write a small test to cover the scenario you described?

codeliner:
Can imagine that it is quite hard to provide a test case for this scenario. What do you think @sandrokeil ?

kochen:
Something like:
1. given a state (after first run)
2. measure memory usage (representing the 320 MB)
3. run again
4. assert memory usage is not higher than expected (< 2*320MB or even < 512MB)

prolic:
@sandrokeil @codeliner want me to push and release this? or you still waiting for test case?

sandrokeil:
We can create an issue to add a test for this later. @prolic I'm for merge it now and create a bug fix release. What do you think @codeliner ?

codeliner:
sorry for being late. I'm ok with a release. Go for it @prolic

PR #235, opened 2021-12-09
Fix MysqlEventStore checking if stream exists for php8.1

dannydorfel:
The return of the pdo statement when executing `$statement->fetchColumn()` in php8.1 is an integer. So the `'1' === $statement->fetchColumn();` will fail. 

Since we check if we got 1 or 0, which is always a number, casting to int makes it a more secure check.


prolic:
thanks!

PR #236, opened 2022-02-14
Ensure PHP 8.1 compatibility

tomcizek:


enumag:
Looks good but CI didn't trigger... any ideas why @prolic? :thinking:

PR #239, opened 2022-05-17
Allow int return values from GET_LOCK MySQL function

rmbl:
After updating our application to PHP 8.1 and updating our dependencies, all functions that tried to save our event store models failed because of a failure to get a lock. It seems like either the PHP update or an update of a doctrine library changed the return value of the GET_LOCK MySQL function to return int instead of a string.


prolic:
We don't use doctrine :-p

rmbl:
shit. true :)

guess it was the update to php 8.1 then

prolic:
released https://github.com/prooph/pdo-event-store/releases/tag/v1.14.1

thank you!

rmbl:
thank you for the quick review and release!

PR #242, opened 2022-09-07
enable to set amount of event being loaded in single run

unixslayer:
with large amount of events to load being able to load only part of it improves performance on reloading projection


prolic (approved):

prolic:
thanks @unixslayer

unixslayer:
@prolic can you run a release with it?

prolic:
@unixslayer done https://github.com/prooph/pdo-event-store/releases/tag/v1.15.0

Issue #243, opened 2022-09-28
Gapdetection default wrong time definition


hvanoch:
https://github.com/prooph/pdo-event-store/blob/541fc4dc28f98d6d6d94f4f18d8dd00da4d267cf/src/Projection/GapDetection.php#L39 defines the time to sleep in ms. But when the sleep is actually called, a `usleep` is used. 
https://github.com/prooph/pdo-event-store/blob/541fc4dc28f98d6d6d94f4f18d8dd00da4d267cf/src/Projection/PdoEventStoreReadModelProjector.php#L519

So this should be multiplied by a factor 1000

prolic:
@hvanoch do you want to submit a pull request and update the docs?

DPDmancul:
Agreeing on the fact that multiplying all times by 1000 (as done by #244) is a big breaking change, I have done two alternative PRs:
- #262 fixes only the documentation
- #263 fixes the documentation and the default times (bringing the effective default to the intended one)

I suggest merging the second, since the current effective default (up to 500 μs) is too low and most users do not customize the gap detection times, relying on the default values.

PR #244, opened 2022-10-20
fix gap detection

prolic:
resolves https://github.com/prooph/pdo-event-store/issues/243

/cc @hvanoch

### DIFF
--- a/src/Projection/PdoEventStoreProjector.php
+++ b/src/Projection/PdoEventStoreProjector.php
@@ -551,16 +551,14 @@ EOT;
                 }
 
                 if ($gapDetected && $this->gapDetection) {
-                    $sleep = $this->gapDetection->getSleepForNextRetry();
-
-                    \usleep($sleep);
+                    \usleep($this->gapDetection->getSleepForNextRetry() * 1000);
                     $this->gapDetection->trackRetry();
                     $this->persist();
                 } else {
                     $this->gapDetection && $this->gapDetection->resetRetries();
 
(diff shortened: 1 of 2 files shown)


src/Projection/PdoEventStoreProjector.php
```diff
@@ -551,16 +551,14 @@ public function run(bool $keepRunning = true): void
                 }
 
                 if ($gapDetected && $this->gapDetection) {
-                    $sleep = $this->gapDetection->getSleepForNextRetry();
-
-                    \usleep($sleep);
+                    \usleep($this->gapDetection->getSleepForNextRetry() * 1000);
```
codeliner, line 554: I'm thinking about whether we should do the calculation or change the documentation and tell people that sleep time is configured in microseconds :thinking: 

When multiplying sleep time 1 ms is the shortest sleep interval possible. Maybe that's too long for some use cases?

codeliner (changes requested):

src/Projection/PdoEventStoreProjector.php
codeliner: The good thing is, changing documentation wouldn't change the current behavior. Gap detection is used in production systems, so people have already identified the best sleep time settings for their projections.

fritz-gerneth: I'm with @codeliner here - afaik this is a BC (as unfortunate it may be).

prolic:
Agreed.

PR #245, opened 2022-11-25
set to IDLE when no events loaded

unixslayer:
this will let client know that projection may need to be run once more if events are loaded in limited batch

### DIFF
--- a/src/Projection/PdoEventStoreReadModelProjector.php
+++ b/src/Projection/PdoEventStoreReadModelProjector.php
@@ -132,6 +132,11 @@ final class PdoEventStoreReadModelProjector implements ReadModelProjector
     /**
      * @var int
      */
+    private $loadedEvents = 0;
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Projection/PdoEventStoreReadModelProjector.php
```diff
@@ -889,8 +895,14 @@ private function releaseLock(): void
 EOT;
 
         $statement = $this->connection->prepare($sql);
+
+        $status = ProjectionStatus::IDLE();
```
prolic, line 899:
```
$status = $this->loadedEvents > 0
    ? ProjectionStatus::RUNNING()
    : ProjectionStatus::IDLE();
```

unixslayer: @prolic updated. if looks ok, please release after merge, thanks ;)

prolic (approved):

PR #247, opened 2025-03-04
remove php8.4 deprecation warnings

zluiten:
Follow-up of https://github.com/prooph/event-store/pull/435


kochen:
@prolic could this be released?

PR #249, opened 2025-03-06
move to github actions

### LINKED ISSUE #248: Remove travis CI builds and switch to github actions
@basz or @zluiten maybe wanna jump in?

zluiten:
Closes #248


zluiten:
This repository uses tests classes from https://github.com/prooph/event-store/pull/435 as their base class, but that doesn't work anymore since - probably for a long time already - since development classes are not installed anymore when adding dependencies.

For now I've worked around it by git cloning the https://github.com/prooph/event-store 7.x -branch as a separate step and adjusted the composer autoloading location.

The proper way might be to extract the base test classes from https://github.com/prooph/event-store into a separate dev repository or just copy the classes in this repo.

src/WriteLockStrategy/MariaDbMetadataLockStrategy.php
```diff
@@ -55,7 +55,7 @@ public function getLock(string $name): bool
         }
 
         $lockStatus = $res->fetchAll();
-        if ('1' === $lockStatus[0]['get_lock']) {
+        if ('1' === $lockStatus[0]['get_lock'] || 1 === $lockStatus[0]['get_lock']) {
```
zluiten, line 58: Fixes failing test and aligns with `\Prooph\EventStore\Pdo\WriteLockStrategy\MysqlMetadataLockStrategy`

tests/MariaDbEventStoreTestCase.php
```diff
@@ -62,7 +62,7 @@ protected function setUp(): void
 
     /**
      * @test
-     * @medium
+     * @large
```
zluiten, line 65: Gives this test more time to succeed. @medium gives 10 seconds, but is sometimes too short

zluiten:
Build in the fork https://github.com/zluiten/pdo-event-store/actions/runs/13721631181

zluiten:
@prolic It's ready for review

Specifically, what are your thoughts regarding this comment? https://github.com/prooph/pdo-event-store/pull/249#issuecomment-2705810751

prolic:
> This repository uses tests classes from [prooph/event-store#435](https://github.com/prooph/event-store/pull/435) as their base class, but that doesn't work anymore - probably for a long time already - since development classes are not installed anymore when adding dependencies.
> 
> For now I've worked around it by git cloning the https://github.com/prooph/event-store 7.x -branch as a separate step and adjusted the composer autoloading location.
> 
> The proper way might be to extract the base test classes from https://github.com/prooph/event-store into a separate dev repository or just copy the classes in this repo.

What if we `require-dev` prooph/event-store? Does that work?

zluiten:
> What if we `require-dev` prooph/event-store? Does that work?

It doesn't unfortunately. The behavior of composer is that test classes referenced in the `autoload-dev` section are excluded from files that are installed when added as a dependency. At least, this I deducted from the `autoload-dev` section in the docs: https://getcomposer.org/doc/04-schema.md#autoload-dev

prolic:
@basz @codeliner suggestions?

basz:
not really no. Seems a separate dev repository is indeed the proper way. But I'm not sure it's worth it at this point. 

A solution might be to modify the event-store repository to separate the base test classes and have those included in the "autoload" (as apposed to "autoload-dev". Meaning production would always have these base test classes.

Or even just move [line](https://github.com/prooph/event-store/blob/master/composer.json#L39C8-L39C18) to [here](https://github.com/prooph/event-store/blob/master/composer.json#L34).

prolic:
I'm good with either.

zluiten:
Done, see https://github.com/prooph/event-store/pull/441

In the end I went with 

```json
    "autoload": {
        "psr-4": {
            "Prooph\\EventStore\\": "src/",
            "ProophTest\\EventStore\\": "tests/"
        },
        "exclude-from-classmap": ["tests/"]
    },
```

The exclude-from-classmap directive makes sure that the test files are excluded from the classmap which is generated when optimizing the autoloader so the autoloader in production doesn't get polluted with these test files (the test files will still be installed though).

prolic:
This looks great. @basz what do you think?

prolic:
https://github.com/prooph/event-store/pull/441 is merged. I'll let @basz take over from here. Feel free to create a release when done.

zluiten:
Thanks @prolic We just need the a new release of prooph/event-store with the new changes for this build to succeed.

basz:
hi @zluiten I just released it as v7.11

that should be placed [here](https://github.com/prooph/pdo-event-store/blob/master/composer.json#L21)

basz:
Not sure what's going on  and maybe you aware. But those test classes can't be found... https://github.com/zluiten/pdo-event-store/actions/runs/14059329730/job/39366037771#step:6:14 

also. @prolic did you mean to give me maintainer status here too? Cause I don't have it...

zluiten:
Shoot..... the test classes are not in the released source files. See also the zip file of the release https://github.com/prooph/event-store/archive/refs/tags/v7.11.0.zip

I tested it with adding my fork of prooph/event-store as a custom repository to the composer.json...

@basz any idea how github decided what is included in the source zip file?

basz:
I'm guessing this

https://github.com/prooph/event-store/blob/master/.gitattributes#L3

basz:
Since we're going this route I think it's more consistent if we include these in the release.

Let's remove that line and i'll do a patch release.

zluiten:
@basz now i wonder if the psr change was needed after all :) maybe it was a matter of removing the line from .gitattributes?

basz:
That _is_ needed also. Or the autoloader still won't work.

basz:
I just released https://github.com/prooph/event-store/releases/tag/v7.11.1

zluiten:
build is ✅

zluiten:
@prolic you are still needed here :) or upgrade @basz to maintainer

prolic:
@basz you should be maintainer now

PR #250, opened 2025-03-29
compatibility with https://github.com/prooph/event-store/pull/442

### LINKED ISSUE #246: Projection not being reset to idle
I think the change in #245 is causing a problem. If the projection is not set to run forever/continuously, the status will stay at running. I've duplicated it by calling `run(false)` on the projection and also using `bin/console event-store:projection:run -o <projection>` in a Symfony app. I'm not quite sure of another option.

unixslayer:
fixes #246

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -59,7 +59,7 @@ jobs:
         name: Coverage Report as Comment (Clover)
         uses: lucassabreu/comment-coverage-clover@main
         with:
-          file: coverage.xml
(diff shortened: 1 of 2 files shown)


basz:
@unixslayer I noticed that this branch has some conflicts now. I just merged an other branch and I'm not sure but that might be why...

unixslayer:
@basz when can this be released?

unixslayer:
@basz I've added also a fix for #246 to this PR

basz:
thanks! its released

PR #251, opened 2025-03-31
Increase lowest required event-store version

unixslayer:


unixslayer:
@basz forgot to increase lowest version for event-store

PR #252, opened 2025-04-02
PHP 8.1 is still supported

unixslayer:


unixslayer:
@basz here is the last one regarding PHP 8.1 support

PR #254, opened 2025-04-02
PHP 8.1 is still supported

unixslayer:


unixslayer:
@basz and the last one. once this is merged, it can be released.

PR #257, opened 2025-04-16
keep projections running as long events are loading

unixslayer:
projections should keep running when:
- runs with `keepRunning: true` due it awaits for new events
- it did not reached end of the stream, e.g. in case of resetting

projection status should indicate actual status of projection therefore, `idle` should be set if projection wasn't supposed to keep running or end of stream was reached

### DIFF
--- a/tests/Projection/AbstractPdoEventStoreReadModelProjectorTestCase.php
+++ b/tests/Projection/AbstractPdoEventStoreReadModelProjectorTestCase.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace ProophTest\EventStore\Pdo\Projection;
 
 use ArrayIterator;
+use Assert\Assertion;
 use DateInterval;
 use DateTimeImmutable;
 use DateTimeZone;
@@ -27,6 +28,7 @@ use Prooph\EventStore\Exception\InvalidArgumentException;
 use Prooph\EventStore\Pdo\Projection\GapDetection;
 use Prooph\EventStore\Pdo\Projection\PdoEventStoreProjector;
 use Prooph\EventStore\Pdo\Projection\PdoEventStoreReadModelProjector;
+use Prooph\EventStore\Projection\ProjectionStatus;
 use Prooph\EventStore\Projection\Projector;
 use Prooph\EventStore\Projection\ReadModel;
 use Prooph\EventStore\Stream;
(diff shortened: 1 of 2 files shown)


unixslayer:
@basz @prolic please, review.

unixslayer:
@prolic done

basz:
I can have a look later today. But am wondering if this should be considered a breaking change release.

Even though it's a small bug fix it changes a side effect that consuming applications might rely on.

Any advise on this?

unixslayer:
> I can have a look later today. But am wondering if this should be considered a breaking change release.
> 
> Even though it's a small bug fix it changes a side effect that consuming applications might rely on.
> 
> Any advise on this?

@basz this is somehow a follow-up to #246.

For projections running with `keepRunning: true` nothing changes. The problem comes with projections with `keepRunning: false`. Especially if you want to reset them. Before that bug was solved, projection status remains `running` with lock released if `load_count` is low so we know it has to be triggered once more so it will go to the end of the stream. But now, projection status is set to `idle` after handling loaded events therefore, before the end of the stream so we don't know if it actually finished.

I don't consider this as a BC but more like a critical issue fix in the context of projection management.

unixslayer:
We are running all our projections with `keepRunning: false`  so we don't have long running processes and trigger them only when they need to be triggered.

basz:
Any tests included inside files named *TestCase aren't executed it seems. 

(see with vendor/bin/phpunit --testdox)

Don't know if this slipped through or is an deliberate choice, but a *lot* of tests never ran...

basz:
see `./vendor/bin/phpunit --list-tests -c phpunit.postgres.xml`

unixslayer:
I didn't even notice that. Normally I run my own tests locally and let the test suite run in CI.

I think this is happening due to changes in PHPUnit itself. We've had that issue in the project some time ago when switching to a newer version.

I can check that later.

unixslayer:
@basz this one is also ready for review and release ;)

tests/Projection/AbstractPdoEventStoreReadModelProjectorTestCase.php
```diff
@@ -456,27 +458,21 @@ public function it_detects_gap_and_performs_retry(): void
...
-            ->run(false);
-
-        $this->assertEquals(1, $projectionManager->fetchProjectionStreamPositions('test_projection')['user']);
-
-        $this->assertTrue($gapDetection->isRetrying());
+            ->whenAny(
+                function (array $state, Message $event) use ($projectionManager, $gapDetection, $parallelConnection): array {
+                    if ($state['iteration'] === 1) {
+                        Assertion::true($gapDetection->isRetrying());
+                        $parallelConnection->commit();
```
unixslayer, line 467: I had to change a little bit logic in this test due now each projection will try to finish by itself

unixslayer:
@basz sorry for pushing but is there any chance to have this reviewed by today?

basz:
nice work. one thing i noticed without the changes to the PdoEventStoreReadModelProjector.php the tests never run to completion. I would expect a failing test in such a case. going to merge anyway, feel free to have a look at that.

PR #258, opened 2025-04-18
PHPUnit compatibility

unixslayer:
- fix `prooph/event-store` BC Breaks
- upgrade to PHPUnit 10
- test classes should be suffixed with Test
- abstract Test classes should be suffixed with TestCase
- remove remaining deprecations

CC @prolic @basz


unixslayer:
this PR should be followed-up after https://github.com/prooph/event-store/pull/449

unixslayer:
@basz PR ready for review

PR #260, opened 2025-06-06
Introduce GapDetector interface to allow custom implementations

zluiten:
This enables actively tracking how often gaps occur, how many retries are needed to fill the gaps and f.e. log a warning when a permanent gap occurs.

Unfortunately that is quite difficult with the current closed GapDetection class. An easy way would be to undo marking the GapDection class as final to allow extending it in user land code. I guess the proper way would be to add a GapDetector interface.

The latter is what I propose in this PR. It should be backwards compatible.

### DIFF
--- a/src/Projection/GapDetection.php
+++ b/src/Projection/GapDetection.php
@@ -25,7 +25,7 @@ use Prooph\Common\Messaging\Message;
  *
  * @package Prooph\EventStore\Pdo\Projection
  */
-final class GapDetection
+final class GapDetection implements GapDetector
 {
     /**
      * If gap still exists after all retries, projector moves on
--- a/src/Projection/GapDetector.php
+++ b/src/Projection/GapDetector.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * This file is part of prooph/pdo-event-store.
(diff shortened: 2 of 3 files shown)


zluiten:
Hmm, the `bc-check` fails. Anyone can I explain why this would be a BC break?

Edit: It looks like this is a false positive: https://github.com/Roave/BackwardCompatibilityCheck/issues/497#issuecomment-2948454646

prolic:
That's because the pdo event store type hints against concrete
implementation instead of interface in its constructor. IMHO that's an
acceptable break, but should be documented.

zluiten:
In this case it's not a bc break at all I believe, because 
- the PdoEventStoreProjector is final and thus cannot be extended ✅ 
- the concrete implementation used as param type gets the interface implemented  ✅ 

I don't think it's even a minor break (unless users do some sorcery with changing the class using reflection during runtime :) )

zluiten:
@basz or @prolic can one of you have a look/review? 🙏

basz:
Sorry didn't see this. Now am ill. I'll put it on my list.

zluiten:
Thanks @basz !

